PR #41, opened 2017-12-12
acc_provision changes for CF

amittbose:
Signed-off-by: Amit Bose <amitbose@gmail.com>

### DIFF
--- a/provision/acc_provision/acc_provision.py
+++ b/provision/acc_provision/acc_provision.py
@@ -57,6 +57,31 @@ VERSIONS = {
     },
 }
 
+# Known Flavor options:
+# - template_generator: Function that generates the output config
+#       file. Default: generate_kube_yaml.
+# - version_fields: List of config options that must be specified for
+#       the specific version of deployment. Default: VERSION_FIELDS.
+# - apic: Dict that is used for configuring ApicKubeConfig
+#       Known sub-options:
+#       - use_kubeapi_vlan: Whether kubeapi_vlan should be used. Default: True.
+#       - tenant_generator: Name of the function to generate tenant objects.
+#             Default: kube_tn.
+#       - associate_aep_to_nested_inside_domain: Whether AEP should be attached
+#             to nested_inside domain. Default: False.
+KubeFlavorOptions = {}
+
+CfFlavorOptions = {
+    'apic': {
+        'use_kubeapi_vlan': False,
+        'tenant_generator': 'cloudfoundry_tn',
+        'associate_aep_to_nested_inside_domain': True,
+    },
+    'version_fields': [],
+}
+
+DEFAULT_FLAVOR_OPTIONS = KubeFlavorOptions
+
 FLAVORS = {
     "kubernetes-1.8": {
         "desc": "Kubernetes 1.8",
@@ -115,6 +140,18 @@ FLAVORS = {
             },
         },
     },
+    "cloudfoundry-1.0": {
+        "desc": "CloudFoundry cf-deployment 1.x",
+        "default_version": "latest",
+        "config": {
+            "aci_config": {
+                "vmm_domain": {
+                    "type": "CloudFoundry",
+                },
+            },
+        },
+        "options": CfFlavorOptions,
+    }
 }
 
 
@@ -136,6 +173,21 @@ def json_indent(s):
 
 def yaml_quote(s):
     return "'%s'" % str(s).replace("'", "''")
+
+
+def yaml_indent(s, **kwargs):
+    return yaml.dump(s, **kwargs)
+
+
+def yaml_list_dict(l):
+    out = "\n"
+    for d in l:
+        keys = sorted(d.keys())
+        prefix = "  - "
+        for k in keys:
+            out += "%s%s: %s\n" % (prefix, k, d[k])
+            prefix = "    "
+    return out
 
 
 def deep_merge(user, default):
@@ -258,6 +310,7 @@ def cidr_split(cidr):
 def config_adjust(args, config, prov_apic, no_random):
     system_id = config["aci_config"]["system_id"]
     infra_vlan = config["net_config"]["infra_vlan"]
+    node_subnet = config["net_config"]["node_subnet"]
     pod_subnet = config["net_config"]["pod_subnet"]
     extern_dynamic = config["net_config"]["extern_dynamic"]
     extern_static = config["net_config"]["extern_static"]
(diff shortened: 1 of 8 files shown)


provision/acc_provision/acc_provision.py
```diff
@@ -148,6 +174,14 @@ def deep_merge(user, default):
     return copy.deepcopy(user)
 
 
+def is_cloudfoundry_flavor(flavor):
```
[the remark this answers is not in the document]
amittbose: ok

provision/acc_provision/apic_provision.py
```diff
@@ -208,6 +216,20 @@ class ApicKubeConfig(object):
     def __init__(self, config):
         self.config = config
 
+    def is_kube(self):
```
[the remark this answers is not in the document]
amittbose: ok

provision/acc_provision/apic_provision.py
```diff
@@ -469,6 +485,21 @@ def nested_dom(self):
...
+                {
+                    "fvnsEncapBlk": {
+                        "attributes": {
+                            "from": "vlan-%d" % kubeapi_vlan,
+                            "to": "vlan-%d" % kubeapi_vlan,
+                        }
+                    }
+                })
+        if self.is_cf():
+            # TODO This should be required for Kubernetes as well
```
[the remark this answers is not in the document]
amittbose: Without this option, the ESX vSwitch drops the reply packets on the service VLAN because the source MAC (node service MAC) is different from that of the VM's interface.

amittbose:
Addressed review comments

provision/acc_provision/acc_provision.py
```diff
@@ -115,8 +115,23 @@
             },
         },
     },
+    "cloudfoundry-1.0": {
+        "desc": "CloudFoundry cf-deployment 1.x",
+        "default_version": "latest",
+        "config": {
```
readams: Shouldn't you need to set something in here to enable CF mode?

The intention here is that any differences between versions are expressed in terms of config options, and the flavors just initialize the default values for all the config options.  For example if kubeapi_vlan isn't used for CF, we have an option "use_kubeapi_vlan" and set to true for Kube and false for CF.  Then the code that requires divergent behavior is expressed in the form of one of these options.

This means that if future versions of CF and kube differ in other ways we can always express this appropriately using the default config in the flavor.

I guess this was not as clear as I thought it might be from the code; maybe there needs to be a comment :-)

provision/acc_provision/acc_provision.py
```diff
@@ -151,6 +181,8 @@ def deep_merge(user, default):
 def config_default():
     # Default values for configuration
     default_config = {
+        "template_type": TEMPLATE_KUBE,
```
readams: should be "aci-containers.yaml" not "kube" right?

provision/acc_provision/acc_provision.py
```diff
@@ -151,6 +181,8 @@ def deep_merge(user, default):
...
+        "template_generator": generate_kube_yaml,
```
readams: Putting a function in here is asking for trouble since the user can set this parameter in their input YAML file, at which point presumably we just blindly call the string as a function.

provision/acc_provision/acc_provision.py
```diff
@@ -573,8 +666,63 @@ def generate_kube_yaml(config, output):
...
+    return config
+
+
 def generate_apic_config(config, prov_apic, apic_file):
-    apic_config = ApicKubeConfig(config).get_config()
+    configurator = ApicKubeConfig(config)
+    if config["template_type"] == TEMPLATE_CF:
+        configurator.with_kubeapi_vlan = False
+        configurator.tenant_generator = configurator.cloudfoundry_tn
+        configurator.associate_aep_to_nested_inside_domain = True
```
readams: This option should be moved to the config and specified in the "flavor" configuration section

provision/acc_provision/acc_provision.py
```diff
@@ -573,8 +666,63 @@ def generate_kube_yaml(config, output):
...
+
+    return config
+
+
 def generate_apic_config(config, prov_apic, apic_file):
-    apic_config = ApicKubeConfig(config).get_config()
+    configurator = ApicKubeConfig(config)
+    if config["template_type"] == TEMPLATE_CF:
+        configurator.with_kubeapi_vlan = False
+        configurator.tenant_generator = configurator.cloudfoundry_tn
```
readams: see above about generator function

provision/acc_provision/acc_provision.py
```diff
@@ -573,8 +666,63 @@ def generate_kube_yaml(config, output):
...
+             "cf_ops.yml -l %s" % applyname)
+
+    return config
+
+
 def generate_apic_config(config, prov_apic, apic_file):
-    apic_config = ApicKubeConfig(config).get_config()
+    configurator = ApicKubeConfig(config)
+    if config["template_type"] == TEMPLATE_CF:
+        configurator.with_kubeapi_vlan = False
```
readams: see above about use_kubeapi_vlan in flavor config

provision/acc_provision/acc_provision.py
```diff
@@ -721,6 +869,9 @@ def provision(args, apic_file, no_random):
             })
 
     deep_merge(config, config_default())
+    if flavor.startswith("cloudfoundry"):
```
readams: same issue with string matching on flavor name

provision/acc_provision/acc_provision.py
```diff
@@ -353,17 +384,67 @@ def config_adjust(args, config, prov_apic, no_random):
...
+
     return adj_config
 
 
-def config_validate(config):
+def config_validate(flavor, config):
     def Raise(exception):
         raise exception
     required = lambda x: True if x else Raise(Exception("Missing option"))
+    required_if_kube = (lambda x:
```
readams: ideally these checks would not be expressed hard-coded based on template type; this is identical to matching on flavor

provision/acc_provision/apic_provision.py
```diff
@@ -526,32 +545,52 @@ def associate_aep(self):
...
+                    }
+                })
 
         base = '/api/mo/uni/infra/attentp-%s' % aep_name
         rsvmm = base + '/rsdomP-[uni/vmmp-%s/dom-%s].json' % (vmm_type, vmm_name)
         rsphy = base + '/rsdomP-[uni/phys-%s].json' % phys_name
         rsfun = base + '/gen-default.json'
-        return path, data, rsvmm, rsphy, rsfun
+
+        if self.associate_aep_to_nested_inside_domain:
```
readams: why is this different for CF?

provision/acc_provision/acc_provision.py
amittbose: That makes things clearer. Instead of adding config options for the different behavior I desire, I'm going to add "flavor options" which will be different from "config". These will not be user configurable, unlike the config options. The intention of these flavor options is to alter the execution flow for different flavors. Example of such options are use_kubeapi_vlan, template_generator_function etc

provision/acc_provision/acc_provision.py
amittbose: Having a different template name is not enough. I think there are enough other differences in generating the output file that it deserves a separate function of its own. My plan is to make the output generation function a "flavor option"

provision/acc_provision/acc_provision.py
amittbose: This generator function is not user configurable, not sure if your earlier comment applies.

provision/acc_provision/apic_provision.py
amittbose: CF VMs (node EPG) actually use the nested-inside domain rather than a physical domain, so we need to associate the AEP with the nested-inside domain.

provision/acc_provision/acc_provision.py
readams: There's already a lot of things in there that are not really things the user is likely to configure except by setting the flavor, such as most of the OpenShift options.  So not really needed to separate it I think.

amittbose: I think a distinction needs to be made between not likely to configure versus must not be able to modify. I think some of the confusion with the flavor stuff stems from overloading the purpose of config.

readams: well most of the config options are already such that they need to be set correctly or the system won't work.  I don't see why cloudfoundry config items should use a separate method.

The key thing to note is that I'd expect, e.g. a future version of OpenShift to require fewer workarounds so an advanced user might actually set config options to override the default flavor to run on a newer version.

readams: If you're proposing a broader refactor that's a different question of course; I'm just saying I'd like to keep them consistent.

amittbose:
New patch posted. Please have a look and let me know what you think.

provision/acc_provision/acc_provision.py
```diff
@@ -57,6 +57,31 @@
...
+# - apic: Dict that is used for configuring ApicKubeConfig
+#       Known sub-options:
+#       - use_kubeapi_vlan: Whether kubeapi_vlan should be used. Default: True.
+#       - tenant_generator: Name of the function to generate tenant objects.
+#             Default: kube_tn.
+#       - associate_aep_to_nested_inside_domain: Whether AEP should be attached
+#             to nested_inside domain. Default: False.
+KubeFlavorOptions = {}
+
+CfFlavorOptions = {
```
readams: I'm completely unable to see why this is supposed to be better than just putting these with all the other options?

PR #42, opened 2017-12-12
Change sample indentation to indicate that nested_inside should be a child of vmm_domain.

ceridwen:


readams:
This was actually already in my other change though slightly different to make mistakes even harder.

PR #47, opened 2017-12-18
Debian packaging scripts

amittbose:
Signed-off-by: Amit Bose <bose@noironetworks.com>


Makefile
```diff
@@ -55,6 +63,37 @@ clean-vendor:
 	rm -rf vendor
 clean: clean-dist clean-vendor
 
+PACKAGE = aci-containers
+VERSION_BASE ?= 1.2.0
```
readams: Where does this version come from?  Is this cloudfoundry version?  Should we put a "cf" or something in there?  The versioning is going to get very confusing with all these branches flying around :-)

amittbose: I was thinking of versioning independent of Kubernetes or CloudFoundry. Every release branch  gets a version of its own. The master branch version is always the latest and it is revised up when a branch is cut. Since we have 2 release branches at this point, I chose 1.2 for the master so that 1.0 and 1.1 may be used for those branches.

readams: except we already have 1.6, 1.7, and 1.8 :-)

amittbose: So I should change this to 1.9 then :-) ? Assuming that you are ok with the what I'm proposing, it would mean that going forward, the aci-container versions won't necessarily mean anything as far Kubernetes versions go

readams: Except we do need release branches that exactly correspond to kubernetes releases.  This may not be true in about a year or so once kubernetes APIs stabilize, but it is true for now.  So I suspect we need at least versioned kubernetes release branches (i.e kube18, kube19) plus cf branches that have perhaps different versions.

amittbose: That should be fine - I'm not suggesting that you change when release branches are created, or how they are named. All I need is a version numbering scheme for the Debian/RPM package that is not inconsistent with the branches (i.e. packages from branches created earlier should have lower version numbers). At this point using the K8S version seems convenient, but that could diverge in the future.

readams: It's fine if we need to change the current scheme.  Are you saying you want to just track the kubernetes version then?  So master would be 1.8 for now?

amittbose: Patch updated to use 1.8 as the version number

PR #49, opened 2018-01-09
Use the default apigroup apps for Kubernetes 1.8 because it seems like the upstream bug is fixed.

ceridwen:
Using the extensions apigroup in sauto is causing aci-containers-controller to crash on Kubernetes 1.8.


readams:
checks fail here so I pushed the equivalent change in a different pr

PR #53, opened 2018-01-26
Add the ability to write a cluster report to stdout

ceridwen:
This is to simplify the process of getting a cluster report for Jenkins.

### DIFF
--- a/cmd/acikubectl/cmd/debug.go
+++ b/cmd/acikubectl/cmd/debug.go
@@ -176,8 +176,7 @@ func clusterReport(cmd *cobra.Command, args []string) {
 			})
 		}
 	}
-
-	outfile, err := os.Create(output)
+	output, outfile, err := getOutfile(output)
 	if err != nil {
 		fmt.Fprintln(os.Stderr, err)
 		return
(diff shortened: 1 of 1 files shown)


cmd/acikubectl/cmd/debug.go
```diff
@@ -176,10 +176,8 @@ func clusterReport(cmd *cobra.Command, args []string) {
 			})
 		}
 	}
-
-	outfile, err := os.Create(output)
+	output, outfile, err := getOutfile(output)
 	if err != nil {
-		fmt.Fprintln(os.Stderr, err)
```
readams: this is actually the right place for this line

cmd/acikubectl/cmd/debug.go
```diff
@@ -219,6 +217,20 @@ func clusterReport(cmd *cobra.Command, args []string) {
...
 }
 
+
+func getOutfile(output string) (string, *os.File, error)  {
+	if output == "-" {
+		return "standard output", os.Stdout, nil
+	} else {
+	        outfile, err := os.Create(output)
+		if err != nil {
+		        fmt.Fprintln(os.Stderr, err)
```
readams: don't output the error here; if you're bubbling it up output the error in the caller.  i.e. just return the error returned by os.Create

ceridwen:
I moved the error print-out back to the outer function.

PR #57, opened 2018-01-30
gofmt fixes

amittbose:
Signed-off-by: Amit Bose <amitbose@gmail.com>


readams:
How did you run this? There's a bunch of changes in there that are not how gofmt formats the code.

amittbose:
This is the command I ran:
gofmt -tabwidth=4 -w=true <path>

readams:
Just run with default settings no tabwidth=4

amittbose:
Updated.

PR #66, opened 2018-02-16
Acc provisioning changes for Ipv6:

abhijitherekar:
1. Create ND prefix policy with auto-conf turned off
2. In icmp contract create a icmpv6 filter entry


abhijitherekar:
Need to make more changes will upload one more set when it is ready.

Closing this pull request for now.

PR #67, opened 2018-02-16
Acc provisioning changes for Ipv6:

abhijitherekar:
1. Create ND prefix policy with auto-conf turned off
2. In icmp contract create a icmpv6 filter entry

### DIFF
--- a/provision/acc_provision/apic_provision.py
+++ b/provision/acc_provision/apic_provision.py
@@ -5,6 +5,7 @@ import sys
 
 import requests
 import urllib3
+import ipaddress
 
 urllib3.disable_warnings(urllib3.exceptions.InsecureRequestWarning)
 apic_debug = False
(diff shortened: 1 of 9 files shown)


provision/acc_provision/apic_provision.py
```diff
@@ -1022,6 +1032,14 @@ def kube_tn(self):
                                     }
                                 }
                             ]
+                        },
+                        "ndPfxPol": {
+                            "attributes": {
+                                "ctrl": "on-link,router-address",
+                                "lifetime": "2592000",
+                                "name": "kube-nd-%s" % system_id,
```
readams: Naming convention for these is e.g.:
l3out_rsprov_name = "%s-l3out-allow-all" % system_id
Except that this is needs to be done for things in the common tenant.  For stuff in the "system_id" tenant like this you can just use a sensible fixed name.  Probably something like simple and obvious like "kube-nd-ra-policy".  Also, only add for ipv6 subnets; don't change for ipv4 subnets.

abhijitherekar: Done, I have uploaded the new changes.

PR #71, opened 2018-02-21
ACI-containers Controller changes for IPv6

abhijitherekar:
Changes required to make the POD network work for Ipv6

### DIFF
--- a/pkg/controller/nodes.go
+++ b/pkg/controller/nodes.go
@@ -352,29 +352,45 @@ func (cont *AciController) mergePodNet(podnet *nodePodNetMeta, existingAnnotatio
 	cont.recomputePodNetAnnotation(podnet)
 }
 
-// must have index lock
-func (cont *AciController) checkNodePodNet(nodename string) {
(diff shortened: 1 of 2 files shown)


abhijitherekar:
Made the changes and uploaded a new set.

pkg/controller/nodes.go
```diff
@@ -352,28 +352,47 @@ func (cont *AciController) mergePodNet(podnet *nodePodNetMeta, existingAnnotatio
...
+	return changed
+}
 
+// must have index lock
+func (cont *AciController) checkNodePodNet(nodename string) {
+	changed := false
+	if podnet, ok := cont.nodePodNetCache[nodename]; ok {
+		if !cont.podNetworkIps.V4.Empty() {
+			changed = cont.allocateIpChunk(podnet, true)
+		} else {
```
readams: Shouldn't it allocate ranges for both v4 and v6 if it can?  This seems to be either/or.

abhijitherekar: Yeah, eventually, it has to be taken care. Making the changes accordingly. Uploading a new set.

PR #78, opened 2018-02-27
Enable password to be passed in via environment

delapsley:
* Enables user to pass ACI password in via the
  ACC_PROVISION_PASS environment variable.

### DIFF
--- a/provision/README.md
+++ b/provision/README.md
@@ -22,6 +22,6 @@ optional arguments:
   -u , --username   APIC admin username to use for APIC API access
   -p , --password   APIC admin password to use for APIC API access
   -v, --verbose     Enable debug
-  
-Passw
+
+Password can also be set via the ACC_PROVISION_PASS environment variable.
 ```
--- a/provision/README.md
+++ b/provision/README.md
@@ -22,5 +22,6 @@ optional arguments:
   -u , --username   APIC admin username to use for APIC API access
   -p , --password   APIC admin password to use for APIC API access
   -v, --verbose     Enable debug
-
+  
+Passw
 ```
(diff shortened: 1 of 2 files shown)


provision/README.md
```diff
@@ -22,5 +22,6 @@ optional arguments:
   -u , --username   APIC admin username to use for APIC API access
   -p , --password   APIC admin password to use for APIC API access
   -v, --verbose     Enable debug
-
+  
+Passw
```
readams: Looks like some extraneous content here.

provision/acc_provision/acc_provision.py
```diff
@@ -911,8 +912,12 @@ def provision(args, apic_file, no_random):
...
     if args.username:
         config["aci_config"]["apic_login"]["username"] = args.username
-    if args.password:
-        config["aci_config"]["apic_login"]["password"] = args.password
+
+    try:
+        config["aci_config"]["apic_login"]["password"] = os.environ['ACC_PROVISION_PASS']
+    except KeyError:
+        if args.password:
+            config["aci_config"]["apic_login"]["password"] = args.password
```
readams: This whole chunk of code can simplified to:

    config["aci_config"]["apic_login"]["password"] = \
            os.environ.get('ACC_PROVISION_PASS', args.password)

mandeepdhami: Should the command line parameter not override the one  in environment (that was a problem with the original patch as well)? I would recommend that it be something like:
'''
config["aci_config"]["apic_login"]["password"] = \
    args.password if args.password else os.environ.get('ACC_PROVISION_PASS')
'''

delapsley: Much nicer thanks! Update following shortly.

delapsley:
Thanks @readams. I've updated the comment in the README.

PR #93, opened 2018-03-12
Handle dhcp relay label

kahou82:
There is a situation that the baremetal or VM behind the kube-node-bd
relies on an external dhcp server. This patchset is to allow user
to create the dhcp relay policy label to handle such use case.

### DIFF
--- a/provision/acc_provision/apic_provision.py
+++ b/provision/acc_provision/apic_provision.py
@@ -1370,22 +1370,24 @@ class ApicKubeConfig(object):
 
         # If dhcp_relay_label is present, attach the label to the kube-node-bd
         if "dhcp_relay_label" in self.config["aci_config"]:
-            # lookup kube-node-bd data
             dbg("Handle DHCP Relay Label")
             children = data['fvTenant']['children']
-            for child in children:
-                if ('fvBD' in child and
-                    child['fvBD']['attributes']['name'] == 'kube-node-bd'):
-                    attr = {
-                        "dhcpLbl": {
-                            "attributes": {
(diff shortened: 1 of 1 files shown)


provision/acc_provision/apic_provision.py
```diff
@@ -1303,6 +1303,24 @@ def kube_tn(self):
...
+                    child['fvBD']['attributes']['name'] == 'kube-node-bd'):
+                    attr = {
+                        "dhcpLbl": {
+                            "attributes": {
+                                "name": self.config["aci_config"]["dhcp_relay_label"],
+                                "owner": "infra"
+                            }
+                        }
+                    }
+                    child['fvBD']["children"].append(attr)
```
mandeepdhami: Slight refactor can help readability like:
```python
        if "dhcp_relay_label" in self.config["aci_config"]:
            dbg("Handle DHCP Relay Label")
            children = data['fvTenant']['children']
            dhcp_relay_label = self.config["aci_config"]["dhcp_relay_label"]
            dhcp_attr = {
                 "dhcpLbl": {
                     "attributes": {
                         "name": dhcp_relay_label,
                         "owner": "infra"
                     }
                 }
             }

            # lookup kube-node-bd data
            for child in children:
                if 'fvBD' in child:
                    if child['fvBD']['attributes']['name'] == 'kube-node-bd':
                        child['fvBD']["children"].append(dhcp_attr)
                        dbg("Adding DHCP Relay Label")
                        break
```

mandeepdhami:
1. Please resolve the merge conflict
2. See the refactor comment in-line

PR #96, opened 2018-03-15
Following changes are specific to Cloudfoundry:

puppet-py:
- Added support for creating user defined custom-epgs
  in APIC, needed for annotations
- Added support for creating APIC objects needed for
  isolated-segment enabled CF deployment

### DIFF
--- a/provision/acc_provision/apic_provision.py
+++ b/provision/acc_provision/apic_provision.py
@@ -593,7 +593,8 @@ class ApicKubeConfig(object):
                 })
 
         base = '/api/mo/uni/infra/attentp-%s' % aep_name
-        rsvmm = base + '/rsdomP-[uni/vmmp-%s/dom-%s].json' % (vmm_type, vmm_name)
+        rsvmm = base + \
+            '/rsdomP-[uni/vmmp-%s/dom-%s].json' % (vmm_type, vmm_name)
         rsphy = base + '/rsdomP-[uni/phys-%s].json' % phys_name
 
         if self.associate_aep_to_nested_inside_domain:
@@ -1441,7 +1442,7 @@ class ApicKubeConfig(object):
         vmm_name = self.config["aci_config"]["vmm_domain"]["domain"]
         cf_vrf = self.config["aci_config"]["vrf"]["name"]
         cf_l3out = self.config["aci_config"]["l3out"]["name"]
-        node_subnet = self.config["net_config"]["node_subnet"]
+        node_subnet = [self.config["net_config"]["node_subnet"]]
         pod_subnet = self.config["net_config"]["pod_subnet"]
         vmm_type = self.config["aci_config"]["vmm_domain"]["type"]
         nvmm_name = (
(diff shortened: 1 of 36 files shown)


provision/acc_provision/apic_provision.py
```diff
@@ -1431,6 +1431,37 @@ def cloudfoundry_tn(self):
                    ["app_profile"])
         app_epg_name = (
             self.config["cf_config"]["default_endpoint_group"]["group"])
+        custom_epg_name = (
+            self.config["aci_config"]["cloudfoundry"]["custom_epg"])
+        isolated_segment = (
+            self.config["aci_config"]["cloudfoundry"]["isolated_segment"])
+        children_epg_objs, children_apic_objects = [], []
+
+        if isolated_segment:
```
amittbose: Looks like you are assuming a single isolation segment, but we need to support multiple.

provision/acc_provision/apic_provision.py
```diff
@@ -1431,6 +1431,37 @@ def cloudfoundry_tn(self):
...
+            isolated_subnet = (
+            self.config["aci_config"]["cloudfoundry"]["isolated_subnet"])
```
amittbose: indentation is wrong here and throughout this patch. Please fix all pep8 issues

provision/acc_provision/apic_provision.py
```diff
@@ -1431,6 +1431,37 @@ def cloudfoundry_tn(self):
...
+            node_subnet.append(isolated_subnet)
+
+            is_gorouter_contract = self.contract(
```
amittbose: we need a separate gorouter-contract for each isolation segment, and the so the name of the IS should be part of the contract name

provision/acc_provision/apic_provision.py
```diff
@@ -1431,6 +1431,37 @@ def cloudfoundry_tn(self):
...
+
+            isolated_node_epg = self.epg(
+            "cf-node-isolated",
+            "cf-node-bd",
+            provides=["is-node-all"],
+            consumes=["is-node-all", "is-gorouter", "%s-l3out-allow-all" % system_id],
+            vmm_domains=[(nvmm_type, nvmm_name)])
+
+            isolated_app_epg = self.epg(
+            "cf-isolated-app",
```
amittbose: the name of the app-epg for isolation segment is the same as the name of the isolation segment (and that name needs to be a config parameter)

provision/acc_provision/apic_provision.py
```diff
@@ -1441,19 +1472,31 @@ def cloudfoundry_tn(self):
...
-            consumes=["gorouter", "%s-l3out-allow-all" % system_id],
+            provides=["dns","is-node-all"],
+            consumes=["is-node-all", "gorouter",
+                      "%s-l3out-allow-all" % system_id],
             vmm_domains=[(nvmm_type, nvmm_name)])
+        children_epg_objs.extend((node_epg, app_default_epg))
+        if custom_epg_name:
+           custom_epg = []
+           for name in custom_epg_name:
+               custom_epg.append(self.epg(name,
```
amittbose: why not call children_epg_objs.extend() directly here?

provision/acc_provision/templates/provision-config.yaml
```diff
@@ -2,21 +2,29 @@
 # Configuration for ACI Fabric
 #
 aci_config:
-  system_id: mykube             # Every opflex cluster must have a distict ID
+  system_id: mykube              # Every opflex cluster must have a distict ID
```
amittbose: there seem to quite a few unnecessary/unrelated changes in this (spacing, comments etc), can you remove them? This is a sample file, so for your testing please create a separate input config file instead of modifying this one.

provision/acc_provision/templates/provision-config.yaml
```diff
@@ -2,21 +2,29 @@
...
-      end: 225.20.255.255
-    nested_inside:              # Include if nested inside a VMM; required for CloudFoundry
-    # type: vmware              # Specify the VMM vendor (supported: vmware)
-    # name: myvmware            # Specify the name of the VMM domain
+      start: 225.21.1.1
+      end: 225.21.255.255
+    nested_inside:
+    #  type: vmware
+    #  name: myvmware
+  cloudfoundry:                 # CloudFoundry specific config parameters
```
amittbose: this section should be commented out in the sample config file. The comment should also  mention that this is an optional section.

provision/acc_provision/templates/provision-config.yaml
```diff
@@ -2,21 +2,29 @@
...
+    custom_epg: ["cf-group1", "cf-group2"]
+    isolated_segment: True
```
amittbose: Need to support multiple IS. I think a list of dicts like this is needed:
cloudfoundry:
[..]
    isolated_segments:
        - name: foo                           # required
           subnet: 1.2.3.4/28             # required
        - name: ...
           subnet: ...
Also, these values need to be validated in config_validate()

readams:
This PR is a mess and seems to have somehow included a bunch of differently-based patches from master.

PR #100, opened 2018-03-28

abhijitherekar:


pkg/controller/network_policy.go
```diff
@@ -205,6 +205,9 @@ func (cont *AciController) staticNetPolObjs() apicapi.ApicSlice {
 			outbound := apicapi.NewHostprotRule(ingressSubj.GetDn(),
 				"allow-all-reflexive")
 			outbound.SetAttr("direction", "ingress")
+			if !cont.configuredPodNetworkIps.V6.Empty() {
+				outbound.SetAttr("ethertype", "ipv6")
+			}
 			outbound.SetAttr("ethertype", "ipv4")
```
readams: This does nothing as it immediately overwrites the value with "ipv4".  All these changes have the same problem.  Need to create a separate rule that allows ipv6.

PR #103, opened 2018-03-28

abhijitherekar:


pkg/controller/network_policy.go
```diff
@@ -202,11 +202,20 @@ func (cont *AciController) staticNetPolObjs() apicapi.ApicSlice {
 	{
 		ingressSubj := apicapi.NewHostprotSubj(hppIngress.GetDn(), "ingress")
 		{
-			outbound := apicapi.NewHostprotRule(ingressSubj.GetDn(),
-				"allow-all-reflexive")
-			outbound.SetAttr("direction", "ingress")
-			outbound.SetAttr("ethertype", "ipv4")
-			ingressSubj.AddChild(outbound)
+			if !cont.configuredPodNetworkIps.V6.Empty() {
```
readams: there's no need to do the extra level of indentation here since you're adding the if statements.  These extra indents are just to scope the variables to make sure they don't get accidentally reused further down.  Also, I'd make the name "allow-all-reflexive-v6" so that the rules will sort properly in the UI.

PR #104, opened 2018-03-29

abhijitherekar:


pkg/controller/network_policy.go
```diff
@@ -202,11 +202,20 @@ func (cont *AciController) staticNetPolObjs() apicapi.ApicSlice {
...
+			ingressSubj.AddChild(outbound)
+			}
+			if !cont.configuredPodNetworkIps.V4.Empty() {
 			outbound := apicapi.NewHostprotRule(ingressSubj.GetDn(),
 				"allow-all-reflexive")
 			outbound.SetAttr("direction", "ingress")
 			outbound.SetAttr("ethertype", "ipv4")
 			ingressSubj.AddChild(outbound)
+            }
 		}
```
readams: Formatting is completely messed up here.  Always run gofmt before checking in

abhijitherekar: Done, somehow the go plugin for my vim has stopped working so it didn't do go fmt, will do manually and upload.

PR #107, opened 2018-04-06
Set the service endpoint using the opflexODev mac value if available.

readams:
Signed-off-by: Rob Adams <readams@readams.net>


pkg/controller/cf_environment.go
```diff
@@ -655,12 +659,12 @@ func (env *CfEnvironment) LoadCellServiceInfo(cellId string) bool {
 			return false
 		}
 	} else {
-		err = json.Unmarshal([]byte(resp.Node.Value), &nodeMeta.serviceEp)
+		err = json.Unmarshal([]byte(resp.Node.Value), &existing)
```
amittbose: 'existing' is a pointer, so you need to drop '&'

PR #109, opened 2018-04-12
Add ability to pass a Kubernetes context argument to acikubectl

ceridwen:
I'm using current context in the test automation.

### DIFF
--- a/cmd/acikubectl/cmd/root.go
+++ b/cmd/acikubectl/cmd/root.go
@@ -29,6 +29,7 @@ import (
 )
 
 var kubeconfig string
+var context string
 
 // RootCmd represents the base command when called without any subcommands
 var RootCmd = &cobra.Command{
@@ -63,6 +64,9 @@ func init() {
 	RootCmd.PersistentFlags().StringVar(&kubeconfig,
 		"kubeconfig", defaultkubeconfig,
 		"Path to the kubeconfig file to use for CLI requests.")
+	RootCmd.PersistentFlags().StringVar(&context,
+		"context", "",
+		"Kubernetes context to use for CLI requests.")
 }
 
 // initConfig reads in config file and ENV variables if set.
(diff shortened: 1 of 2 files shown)


cmd/acikubectl/cmd/root.go
```diff
@@ -76,9 +80,23 @@ func initClient() (kubernetes.Interface, error) {
...
+			restconfig, err = clientcmd.NewNonInteractiveDeferredLoadingClientConfig(
+				&clientcmd.ClientConfigLoadingRules{
+					ExplicitPath: kubeconfig,
+				},
+				&clientcmd.ConfigOverrides{
+					CurrentContext: context,
+				}).ClientConfig()
+			if err != nil {
+				return nil, err
+			}
```
readams: It's a bit silly to have both calls to BuildConfigFromFlags and the explicit NewNonInteractiveDeferredLoadingClientConfig, given that BuildConfigFromFlags just calls NewNonInteractiveDeferredLoadingClientConfig and passes in the kubeconfig.  You can simplify this all to just the "else" branch above.

readams: (See https://github.com/kubernetes/client-go/blob/master/tools/clientcmd/client_config.go#L551)

ceridwen: What happens if I pass a kubeconfig without a context, though?  I'd assumed that the Kubernetes client would then interpret the default empty string as the context and error out.  I didn't see a way around needing to pass `&clientcmd.ConfigOverrides` in one case and nothing in the other case.

readams: default constructed value of context is going to be ""

ceridwen:
Removed the superfluous branch and added a small Dockerfile for building acikubectl.

readams:
Is there something that this dockerfile does that cannot be done with make dist-static/acikubectl  ?

ceridwen:
It lets me build a container that has acikubectl in it, that's all it does.

readams:
Along with a few hundred megabytes of build dependencies... :-)
Can you use the acikubectl that's packaged along with acc-provision?  This should run anywhere.  Using acikubectl in a container is a bit awkward since you'd need to mount your kubeconfig in the container, which is why we never packaged it that way.  How do you use it (maybe there's an easy way I didn't realize?)

If we really need this container to be built, we should build it the same way the other containers are built with a make container-XXX rule and just use the built dist-static binary.  Then you get a very small container.  Best example to work from is the make container-simpleservice rule.

ceridwen:
I'm using this container as input for another container that just takes the acikubectl binary without all of the dependencies, using multistage builds, so the final container is appropriately small.  I am actually accessing a kubeconfig inside a container on Jenkins.  This isn't an ideal solution, but is the best I've worked out so far without spending a bunch more time on it.  Unfortunately, Docker sucks, so there are exactly no clean ways to avoid needing to pull the entire source tree of this repository from the web every time I build my container and no way to include the Dockerfile in my repository.  (I checked.)

The only pre-built packages for acc-provision are for Debian or Red Hat, right?  Will that run on alpine, and if so, how much of a pain will it be to extract that from the packages?  How large are those packages?

readams:
To get a binary that you can use in your container, the command you want is "make dist-static/acikubectl"

readams:
Could you update this PR without that dockerfile and I can merge it.

ceridwen:
Updated.

PR #136, opened 2018-06-15
Added validation for system ID

mandeepdhami:


provision/acc_provision/acc_provision.py
```diff
@@ -540,11 +540,15 @@ def Raise(exception):
              else Raise(Exception("Invalid value: %s; "
                                   "Expected one of: {%s}" %
                                   (x, ','.join(y)))))))
+    isname = lambda x: (1 < len(x) < 32) & \
```
amittbose: did you really mean to use "and" instead of "&"? Also the exception raised should be say that the name is invalid instead of saying option is missing. Something like this

lambda x: (
   True if x and len(x) < 32 and x[0].isalpha() and x.replace('_', '').isalnum()
   else Raise(Exception("Invalid name"))

provision/acc_provision/acc_provision.py
```diff
@@ -540,11 +540,15 @@ def Raise(exception):
...
+        x[0].isalpha() & x.replace('_', '').isalnum() \
+        if x else Raise(Exception("Missing option"))
     get = lambda t: functools.reduce(lambda x, y: x and x.get(y), t, config)
 
     checks = {
         # ACI config
         "aci_config/system_id": (get(("aci_config", "system_id")), required),
```
amittbose: I think you could just change this check instead of adding another one that doesn't correspond to any field in the config.

"aci_config/system_id": (get(("aci_config", "system_id")), lambda x: required(x) and isname(x)),

provision/acc_provision/acc_provision.py
mandeepdhami: Fixed that.

provision/acc_provision/acc_provision.py
mandeepdhami: Got it. I mis-understood the last feedback. Updated now.

mandeepdhami:
I will squash the changes and make a new pull request

PR #138, opened 2018-06-18
Update acc_provision to Python 3 and fix latent bugs

ceridwen:
I apologize for the size of this monster pull request.  I intended to "only" update acc-provision to be compatible with Python 3 as well as Python 2, but unfortunately I found some other skeletons along the way.  Major changes:

* The json stdlib module in 2.7 has a [bug](https://bugs.python.org/issue16333) that causes it to add trailing whitespace when called with the `indent` keyword.  This is fixed in 3.4+ but not 2.7, so I had to add a workaround to fix it in 2.7, and as a consequence many of the lines for the tests had to change to remove the trailing whitespace.  This accounts for many lines of changes.

* Python 2 and 3 seem to order JSON keys differently, probably because of different implementations for dicts, so to get test files that would work for 2 and 3 I had to enforce a deterministic order for the files generated by `acc_provision.py`.  I used `sort_keys` in `json.dumps`, which meant that many more test lines had to change.

* [APIC requires](https://www.cisco.com/c/en/us/td/docs/switches/datacenter/aci/apic/sw/2-x/rest_cfg/2_1_x/b_Cisco_APIC_REST_API_Configuration_Guide/b_Cisco_APIC_REST_API_Configuration_Guide_chapter_01.html) that attributes be the first element of JSON passed to it:
    > "Although the JSON specification allows unordered elements, the APIC REST API requires that the JSON 'attributes' element precede the 'children' array or other elements."

    `apic_provision.py` uses `sort_keys` in `json.dumps` already, and presumably there are no keys that sort before it in the dicts.
    * I've changed all the relevant dicts in `apic_provision.py` to OrderedDicts.  Unfortunately, this has increased the length of code lines and made things harder to read, but preserving insertion order is a complete solution to this problem since you can always insert `attributes` first.
   * I've added asserts that check that `attributes` is first.  These operate only on the OrderedDicts because currently `json.dumps` uses `sort_keys`, so the sorting only occurs *after* the data structures have been serialized.  I'd have to deserialize the data, preserving its order, in order to check that attributes is first after serialization rather than before it.
   * Because `sort_keys` is still in place, adding a key that sorts before `attributes` will still break the APIC API.  I can remove this, but then a bunch of test files need to change.

* I used `from __future__ import unicode_literals` to get consistent string types across 2 and 3, so strings are now unicode by default.

* `os.tempnam` was removed in Python 3, I moved the tests to use `tempfile` instead.

* `filecmp.cmp` does not, AFAICT, actually do a comparison of the *content* of files unless the keyword argument `shallow` is set to false from its default value true.  When `shallow` is true, it only checks the mtime, size, and mode of files to determine if they're identical.  I'm still not quite sure how the tests were passing, but regardless, the files created by`tempfile` usually don't pass the initial mtime/size/mode comparison.  I had to make most of the tests read the generated files and then do a string comparison for validation, though for some of them I could simply set `shallow` to false.

* Some of the tests were erroneously setting sys.stderr to sys.stdout, causing output to get eaten.   I fixed those.

* I fixed a number of smaller Python 2/3 issues and added flushes/seeks to work around buffering.


mandeepdhami:
Difference between dict and OrderedDict in how the code is written makes this difficult to review that change manually. I am assuming that as you have tested against it against the existing output files, the translation is correct. Other than that, LGTM.

PR #145, opened 2018-08-07
Allow `-` in system id

tiewei:
The char `-` was allowed in system id, and keep allowing it here
to make sure old cluster can be upgraded

(cherry picked from commit 5cc8870b5d114c41fe1f660c9a4791238113d632)


mandeepdhami:
We can not allow '-', #146 adds support for '-' when deleting a cluster

PR #148, opened 2018-08-11
Fixed dn for custom portgroup (usrcustomaggr)

mandeepdhami:
Fixed dn for custom portgroup (usrcustomaggr)


mandeepdhami:
Merging as validated by CCP

PR #152, opened 2018-08-24
Added option to set interface MTU for containers

mandeepdhami:
Added option to set interface MTU for containers
Added basic tests for it
Bumped version number

### DIFF
--- a/provision/acc_provision/acc_provision.py
+++ b/provision/acc_provision/acc_provision.py
@@ -292,6 +292,7 @@ def config_default():
             "kubeapi_vlan": None,
             "service_vlan": None,
             "service_monitor_interval": 0,
+            "interface_mtu": None,
         },
         "kube_config": {
             "controller": "1.1.1.1",
@@ -545,6 +546,22 @@ def config_adjust(args, config, prov_apic, no_random):
     return adj_config
 
 
+def is_valid_mtu(xval):
+    if xval is None:
+        # use default configured on thsi host
+        return True
+
+    xmin = 1280   # for IPv6
+    xmax = 8900   # leave 100 byte header for VxLAN
+    try:
+        x = int(xval)
+        if xmin <= x <= xmax:
+            return True
+    except ValueError:
+        pass
+    raise(Exception("Must be integer between %d and %d" % (xmin, xmax)))
+
+
 def config_validate(flavor_opts, config):
     def Raise(exception):
         raise exception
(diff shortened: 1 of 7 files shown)


provision/acc_provision/templates/aci-containers.yaml
```diff
@@ -71,6 +71,9 @@ data:
         "service-vlan": {{ config.net_config.service_vlan|json }},
         "encap-type": {{ config.node_config.encap_type|json }},
         "aci-infra-vlan": {{ config.net_config.infra_vlan|json }},
+        {% if config.net_config.interface_mtu %}
+        "interface-mtu": {{ config.net_config.interface_mtu }},
```
snaiksat: Does this need the '|json" filter as is being used for other configs?

provision/acc_provision/acc_provision.py
```diff
@@ -545,6 +546,22 @@ def config_adjust(args, config, prov_apic, no_random):
     return adj_config
 
 
+def is_valid_mtu(xval):
+    if xval is None:
+        # use default configured on thsi host
```
snaiksat: nit: thsi -> this

provision/acc_provision/acc_provision.py
```diff
@@ -545,6 +546,22 @@ def config_adjust(args, config, prov_apic, no_random):
...
+        return True
+
+    xmin = 1280   # for IPv6
```
snaiksat: Seems like the default we were are using is 1500:
https://github.com/noironetworks/aci-containers/blob/f1a564c9abdf320eaf3af3654bbd5c02c092ce96/pkg/hostagent/config.go#L177

Should xmin be 1500?

provision/acc_provision/templates/aci-containers.yaml
mandeepdhami: Good catch, I will fix that.

provision/acc_provision/acc_provision.py
mandeepdhami: OK ;-)

provision/acc_provision/acc_provision.py
mandeepdhami: No, the user can set it below the default - just not lower than 1280 as IPv6 requires that (IPv4 min is lower, but opflex supports IPv6 so we will go with that min).

PR #156, opened 2018-09-27
Change the APIC refresh Timer to be configurable.

abhijitherekar:
As a part of this change we have a added a new parameter in the
controller configMap.
This parameter will determine the refresh rate at which the the
client with refresh with the APIC.
If no value given then the default is taken.


pkg/apicapi/apicapi.go
```diff
@@ -193,7 +193,7 @@ func New(log *logrus.Logger, apic []string, user string,
 
 	conn := &ApicConnection{
 		ReconnectInterval: time.Second,
-		RefreshInterval:   30 * time.Second,
+		RefreshInterval:   time.Duration(refresh/2) * time.Second,
```
jojimt: Why not just do refresh * time.Second? We shouldn't have to do /2.

abhijitherekar: Just to be pro-active and do the refresh before the APIC time's out.

abhijitherekar: DONE

PR #158, opened 2018-10-05
Ipam fix

jojimt:
Fix for messed up podNet annotations plus the test.

### DIFF
--- a/pkg/controller/nodes.go
+++ b/pkg/controller/nodes.go
@@ -83,7 +83,7 @@ func apicNodeNetPol(name string, tenantName string,
 		inbound.SetAttr("ethertype", "ipv4")
 		inbound.SetAttr("connTrack", "normal")
 
-		for ip, _ := range nodeIps {
+		for ip := range nodeIps {
 			outbound.AddChild(apicapi.NewHostprotRemoteIp(outbound.GetDn(), ip))
 			inbound.AddChild(apicapi.NewHostprotRemoteIp(inbound.GetDn(), ip))
 		}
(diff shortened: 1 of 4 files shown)


pkg/controller/nodes.go
```diff
@@ -83,7 +83,7 @@ func apicNodeNetPol(name string, tenantName string,
 		inbound.SetAttr("ethertype", "ipv4")
 		inbound.SetAttr("connTrack", "normal")
 
-		for ip, _ := range nodeIps {
+		for ip := range nodeIps {
```
mchalla: Is this change necessary ?

pkg/controller/nodes.go
```diff
@@ -251,16 +251,11 @@ func (cont *AciController) nodeChanged(obj interface{}) {
...
 	}
 
-	netval, netok :=
-		node.ObjectMeta.Annotations[metadata.PodNetworkRangeAnnotation]
-	if netok {
-		if netval != nodePodNet.podNetIpsAnnotation {
-			cont.mergePodNet(nodePodNet, netval, logger)
-		}
-	}
+	netval := node.ObjectMeta.Annotations[metadata.PodNetworkRangeAnnotation]
```
mchalla: since its a string map netok should have been true if the annotation was present ? I guess the actual change is that we no longer do the mergePodNet. I need to look at what it does but if you can explain how its causing the overlap issue it would be helpful.

snaiksat:
nit: Can we update the commit message to include an example of the incorrect/redundant annotations we are fixing? It will help anyone looking at the git logs to correlate the issue with the fix.

snaiksat:
I will merge this for now since we need to do a build.

PR #165, opened 2018-10-31
Enhance listing of supported flavors & Docker UCP update

snaiksat:
* Flavors has a 'status' attribute which allows specifying
various options like "Trial", "Pre-release", 'Experimental", etc.
If status attribute is populated, it is appended to the
description of the flavor when the flavors are listed.

* Older supported flavors can be suppressed from the flavor
listing by setting a new "hidden" attribute to True.

* Flavor listing is sorted

* Refactoring of code into separate flavors module

In addition,

* Docker UCP 3.0 flavor points to Kubernetes 1.9 version

### DIFF
--- a/provision/Makefile
+++ b/provision/Makefile
@@ -6,7 +6,7 @@ dist: test
 	python setup.py sdist
 
 test:
-	flake8 --ignore E501,E731,E741 acc_provision
+	flake8 --ignore E501,E731,E741,W504 acc_provision
 	python -m pytest acc_provision
 
 clean:
--- a/provision/acc_provision/acc_provision.py
+++ b/provision/acc_provision/acc_provision.py
@@ -24,7 +24,9 @@ from .apic_provision import Apic, ApicKubeConfig
 from jinja2 import Environment, PackageLoader
 from os.path import exists
 
-DEFAULT_FLAVOR = "kubernetes-1.10"
+from . import flavors
+
+DEFAULT_FLAVOR = flavors.DEFAULT_FLAVOR
 
 VERSION_FIELDS = [
     "cnideploy_version",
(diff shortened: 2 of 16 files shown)


snaiksat:
I should have split this across at least two commits.

mandeepdhami:
It seems that you have un-done the MTU fix. Was that by design?

mandeepdhami:
For flavors, I would recommend that:
1. For K8S:
    a. 1.12 not be hidden, and that it's description be pre-release (not experimental)
    b. 1.9 be hidden
2. For Openshift:
    a. Add OpenShift 3.11 as (not hidden, experimental)

snaiksat:
@mandeepdhami thanks for the review. Good catch on the MTU fix, certainly not by design. Since I was testing this on a Fab, I ended up using the release-1.9 branch codebase and seems like the MTU commit was not back ported to that branch! Will make the other changes as well.

snaiksat:
This is how the listing looks like now:

root@fab10-compute-1:~# acc-provision --list-flavors
INFO: Available configuration flavors:
INFO: cloudfoundry-1.0:	CloudFoundry cf-deployment 1.x
INFO: docker-ucp-3.0:	Docker Universal Control Plane (UCP) 3.0 [Pre-release]
INFO: kubernetes-1.10:	Kubernetes 1.10
INFO: kubernetes-1.11:	Kubernetes 1.11
INFO: kubernetes-1.12:	Kubernetes 1.12 [Pre-release]
INFO: openshift-3.11:	Red Hat OpenShift Container Platform 3.11 [Experimental]
INFO: openshift-3.9:	Red Hat OpenShift Container Platform 3.9

mandeepdhami:
Looks good to me. Merging it.
Minor nit for some future release, we should sort 3.11 after 3.9 ;-)

PR #175, opened 2018-11-08
Multiple fixes related to agent IPAM plus an integration test.

jojimt:
Addresses duplicate IP addresses and other issues related to agent IPAM.

1. Race conditions due to lack of locking in the IPAM
2. Boundary conditions in the IPAM
3. RPC client close to free fds.

Added an integration test to exercise the interactions between CNI and the agent.

### DIFF
--- a/pkg/metadata/cnimetadata.go
+++ b/pkg/metadata/cnimetadata.go
@@ -83,13 +83,13 @@ func LoadMetadata(datadir string, network string,
 	return nil
 }
 
-func CheckMetadata(datadir string, network string) error {
+func CheckMetadata(datadir string, network string) (int64, error) {
 
 	ipMap := make(map[string]string)
 	dir := filepath.Join(datadir, network)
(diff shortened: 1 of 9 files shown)


pkg/metadata/cnimetadata.go
```diff
@@ -49,7 +49,7 @@ type ContainerMetadata struct {
 
 func RecordMetadata(datadir string, network string, data ContainerMetadata) error {
 	dir := filepath.Join(datadir, network)
-	if err := os.MkdirAll(dir, 0644); err != nil {
+	if err := os.MkdirAll(dir, 0755); err != nil {
```
snaiksat: Does this have any impact on upgrades? (as in, if the directory was created with 0644, and then we tried to read it below.)

jojimt: No. This is a bug. I think the only reason this works is that the directory already exists with the right permission (perhaps created by the installer).

PR #179, opened 2018-11-15
Option to add contract scope for k8s services as an annotation

apoorva11029:
Default value: context

For example, add in the service definition:
"metadata": {
  "annotations": {
    "opflex.cisco.com/ext_service_scope" : "tenant"
  }
}

Valid scope values: context, tenant, global. An empty entry defaults to context(VRF only)

If a new service is deployed with invalid scope, it wont register with APIC
If an existing service is updated with invalid scope, the service wont be updated in APIC

Added unit and functional tests to check the contract scope values and service annotations.
Also formatted the committed files with go fmt

### DIFF
--- a/pkg/controller/cf_update_handlers.go
+++ b/pkg/controller/cf_update_handlers.go
@@ -118,7 +118,7 @@ func (env *CfEnvironment) handleContainerUpdateLocked(contId string) bool {
 				Tenant: env.cont.config.AciPolicyTenant})
 		ep.SecurityGroups = append(ep.SecurityGroups,
 			cf_common.GroupInfo{
-				Group: env.cont.aciNameForKey("hpp", "cf-components"),
+				Group:  env.cont.aciNameForKey("hpp", "cf-components"),
 				Tenant: env.cont.config.AciPolicyTenant})
 		if sg != nil {
 			for _, s := range *sg {
@@ -131,22 +131,22 @@ func (env *CfEnvironment) handleContainerUpdateLocked(contId string) bool {
 		if ok {
 			ep.SecurityGroups = append(ep.SecurityGroups,
 				cf_common.GroupInfo{
-					Group: env.cont.aciNameForKey("np", cinfo.AppId),
+					Group:  env.cont.aciNameForKey("np", cinfo.AppId),
 					Tenant: env.cont.config.AciPolicyTenant})
 		}
 		if len(appInfo.ExternalIp) > 0 {
 			ep.SecurityGroups = append(ep.SecurityGroups,
 				cf_common.GroupInfo{
-					Group: env.cont.aciNameForKey("hpp", "app-ext-ip"),
+					Group:  env.cont.aciNameForKey("hpp", "app-ext-ip"),
 					Tenant: env.cont.config.AciPolicyTenant})
 		}
 		if len(env.GetAdditionalPorts(cinfo)) > 0 {
 			ep.SecurityGroups = append(ep.SecurityGroups,
 				cf_common.GroupInfo{
-					Group: env.cont.aciNameForKey("hpp", "app-port:"+cinfo.AppId),
+					Group:  env.cont.aciNameForKey("hpp", "app-port:"+cinfo.AppId),
 					Tenant: env.cont.config.AciPolicyTenant})
 		}
-		env.kvmgr.Set("cell/" + cinfo.CellId, "ct/" + cinfo.ContainerId, &ep)
+		env.kvmgr.Set("cell/"+cinfo.CellId, "ct/"+cinfo.ContainerId, &ep)
 	}
 	if spaceInfo != nil {
 		conf := env.cont.config
@@ -175,7 +175,7 @@ func (env *CfEnvironment) handleContainerDeleteLocked(cinfo *ContainerInfo) bool
 		env.cont.removePodFromNode(cinfo.CellId, cinfo.ContainerId)
 		env.cont.indexMutex.Unlock()
 
-		env.kvmgr.Delete("cell/" + cinfo.CellId, "ct/" + cinfo.ContainerId)
+		env.kvmgr.Delete("cell/"+cinfo.CellId, "ct/"+cinfo.ContainerId)
 	}
 	env.cont.apicConn.ClearApicObjects("inj_contgrp:" + cinfo.ContainerId)
 	return retry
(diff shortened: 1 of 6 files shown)


pkg/metadata/annotations.go
```diff
@@ -46,6 +46,9 @@ type NetIps struct {
 // Service endpoint annotation
 const ServiceEpAnnotation = "opflex.cisco.com/service-endpoint"
 
+// Service scope annotation
+const ServiceScopeAnnotation = "opflex.cisco.com/ext_service_scope"
```
snaiksat: I  am wondering if we should call this ServiceContractScopeAnnotation, and also update the annotation string to:
opflex.cisco.com/ext_service_contract_scope

to make things less ambiguous. Also the comment should be a little more descriptive, indicating what is the relevance of this annotation, how it should be used, what are the legal values, and what is the default. I would recommend reflecting this information in the commit message as well.

pkg/controller/services_test.go
```diff
@@ -185,33 +183,242 @@ func TestServiceIp(t *testing.T) {
...
+	_, err = apicContract(name, tenant, graphName, "wrong-Name")
+	assert.NotNil(t, err)
+
+	contract, err = apicContract(name, tenant, graphName, "")
+	assert.Empty(t, err)
+	assert.Equal(t, contract.GetAttrStr("scope"), "context")
+
+	contract, err = apicContract(name, tenant, graphName, "GloBal")
+	assert.Empty(t, err)
+	assert.Equal(t, contract.GetAttrStr("scope"), "global")
```
snaiksat: Is there an assert needed for the third legal value "tenant"?

pkg/metadata/annotations.go
apoorva11029: Makes sense, I will add this to code and commit.

pkg/controller/services_test.go
apoorva11029: That's to make sure a blank scope value("") defaults to "context"

pkg/controller/services.go
```diff
@@ -325,15 +325,40 @@ func apicExtNetCons(conName string, tenantName string,
...
+	return true
+}
 func apicContract(conName string, tenantName string,
-	graphName string) apicapi.ApicObject {
+	graphName string, scopeName string) (apicapi.ApicObject, error) {
+	scopeName_norm := strings.ToLower(scopeName)
+	checkScope := validScope(scopeName_norm)
+	if !checkScope {
+		return nil, errors.New("InvalidServiceContractScope")
+	}
```
mchalla: could you add the scope name when  returning this error.

pkg/controller/cf_update_handlers.go
```diff
@@ -341,8 +343,11 @@ func (env *CfEnvironment) createAppServiceGraph(appId string, extIps []string) a
 			apicExtNet(name, cont.config.AciVrfTenant,
 				cont.config.AciL3Out, extIps))
 
-		serviceObjs = append(serviceObjs,
-			apicContract(name, cont.config.AciVrfTenant, graphName))
+		contract, err := apicContract(name, cont.config.AciVrfTenant, graphName, conScope)
+		if err != nil {
+			panic(err.Error())
```
jojimt: Let's replace the panic with an error return. Please also open an issue to track the need for validators for this annotation.

pkg/controller/services.go
```diff
@@ -397,11 +422,16 @@ func (cont *AciController) updateServiceDeviceInstance(key string,
 		nodes = append(nodes, node)
 	}
 	sort.Strings(nodes)
-
 	name := cont.aciNameForKey("svc", key)
+	var conScope string
+	scope_val, ok := service.ObjectMeta.Annotations[metadata.ServiceScopeAnnotation]
+	if ok {
+		conScope = scope_val
```
mchalla: I assume ok could also mean its empty "" ?

pkg/controller/services.go
```diff
@@ -325,15 +325,40 @@ func apicExtNetCons(conName string, tenantName string,
...
+		if v == str {
+			return true
+		}
+	}
+	return false
+}
+
+func validScope(scope string) bool {
+	validValues := []string{"", "context", "tenant", "global"}
+	if !stringInSlice(scope, validValues) {
```
jojimt: Please change to return stringInSlice(..) directly.

pkg/controller/services.go
```diff
@@ -325,15 +325,40 @@ func apicExtNetCons(conName string, tenantName string,
...
+		return false
+	}
+	return true
+}
 func apicContract(conName string, tenantName string,
-	graphName string) apicapi.ApicObject {
+	graphName string, scopeName string) (apicapi.ApicObject, error) {
+	scopeName_norm := strings.ToLower(scopeName)
```
jojimt: nit: please stick to just camel case. scopeName_norm can be renamed as normScopeName.

pkg/controller/services.go
jojimt: Can we also avoid declaring checkScope? You can directly check the function return value.

pkg/controller/services.go
apoorva11029: If it's empty, we default to "context", but empty string is still acceptable

pkg/controller/services.go
```diff
@@ -397,11 +422,16 @@ func (cont *AciController) updateServiceDeviceInstance(key string,
...
 	}
 	sort.Strings(nodes)
-
 	name := cont.aciNameForKey("svc", key)
+	var conScope string
+	scope_val, ok := service.ObjectMeta.Annotations[metadata.ServiceScopeAnnotation]
+	if ok {
+		conScope = scope_val
+	} else {
+		conScope = "context"
```
jojimt: Please define a constant named defaultContext and use it here.

pkg/controller/services.go
```diff
@@ -916,7 +951,10 @@ func (cont *AciController) handleServiceUpdate(service *v1.Service) bool {
 		cont.indexMutex.Lock()
 		if cont.serviceSyncEnabled {
 			cont.indexMutex.Unlock()
-			cont.updateServiceDeviceInstance(servicekey, service)
+			err = cont.updateServiceDeviceInstance(servicekey, service)
+			if err != (interface{})(nil) {
```
jojimt: No need for a typecast here. err != nil is sufficient.

pkg/controller/services.go
```diff
@@ -15,25 +15,27 @@
...
+	"sort"
+	"strconv"
+	"strings"
 
 	"github.com/noironetworks/aci-containers/pkg/apicapi"
 	"github.com/noironetworks/aci-containers/pkg/metadata"
 )
 
+var defaultServiceContractScope = "context"
+
```
jojimt: Please define this as const instead of var.

PR #185, opened 2018-11-21
kubernetes 12 nodes come with default taints

mchalla:
Taints:             node-role.kubernetes.io/master:NoSchedule
                    node.kubernetes.io/not-ready:NoSchedule

As a result aci daemonsets cannot start with the current tolerations
- key: CriticalAddonsOnly
- effect: NoSchedule
  key: node-role.kubernetes.io/master

We change this to tolerate any key with effect NoSchedule.

This is the same approach used by weave and other cni plugins.

Signed-off-by: Madhu Challa "challa@noironetworks.com"

### DIFF
--- a/provision/acc_provision/templates/aci-containers.yaml
+++ b/provision/acc_provision/templates/aci-containers.yaml
@@ -281,9 +281,8 @@ spec:
         - name: {{ config.registry.image_pull_secret|yaml_quote }}
       {% endif %}
       tolerations:
-        - key: CriticalAddonsOnly
-        - effect: NoSchedule
-          key: node-role.kubernetes.io/master
+        - operator: Exists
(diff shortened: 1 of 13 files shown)


provision/acc_provision/templates/aci-containers.yaml
```diff
@@ -281,9 +281,8 @@ spec:
         - name: {{ config.registry.image_pull_secret|yaml_quote }}
       {% endif %}
       tolerations:
-        - key: CriticalAddonsOnly
-        - effect: NoSchedule
-          key: node-role.kubernetes.io/master
+        - operator: Exists
+          effect: NoSchedule
```
snaiksat: shouldn't this be:
- effect
(missing '-')

and other places below...

PR #190, opened 2018-12-03
Support for PodIpPoolChunkSize configuration in controller and acc-provision

apoorva11029:
Added code to support PodIpPoolChunkSize configuration in controller and acc-provision. To enable, add the "pod_subnet_chunk_size" field in the input acc provisioning file. Default value: 32

If the PodIpPoolChunkSize is changed:
1. Existing annotations for the pod IP pool on the nodes will not change.
2. The new pod IP pool size will reflect only when a new annotation is applied by the controller after the pool defined by the existing annotation is exhausted.


pkg/controller/config.go
```diff
@@ -134,7 +134,7 @@ type ControllerConfig struct {
 	PodIpPool []ipam.IpRange `json:"pod-ip-pool,omitempty"`
 
 	// The number of IP addresses to allocate when a pod starts to run low
-	PodIpPoolChunkSize int `json:"pod-ip-pool-chunk-size,omitempty"`
+	PodIpPoolChunkSize int `json:"pod-subnet-chunk-size,omitempty"`
```
snaiksat: Was the name changed as a matter of preference or was there a different reason? By changing it to "subnet" it makes it inconsistent with the "ip-pool" terminology used above this and elsewhere.

pkg/controller/config.go
```diff
@@ -180,7 +180,6 @@ func NewConfig() *ControllerConfig {
 		AciVmmDomainType:   "Kubernetes",
```
snaiksat: There should be an info log statement somewhere around here which states what value is actually picked up and being used.

pkg/controller/config.go
snaiksat: Also, what happens if this configuration is not present. I know you have instrumented acc-provision to always populate this but what if the config files gets updated later. There should be a default here as well (if not set).

provision/testdata/base_case.kube.yaml
```diff
@@ -59,6 +59,7 @@ data:
                 "start": "10.2.0.2"
             }
         ],
+        "pod-subnet-chunk-size": 32,
```
snaiksat: We should ideally be testing, setting this to a non-default value, and in another case where nothing is set, and the default gets picked up.

pkg/controller/config.go
apoorva11029: Included in latest commit

apoorva11029: Used this to remain consistent with the name in acc_provision.py

pkg/controller/config.go
apoorva11029: Added in controller.go

provision/testdata/base_case.kube.yaml
apoorva11029: Testing a value set externally in testdata/with_overrides.kube.yaml

FrederikSuijs:
Hi,
kudos to the team :+1: 
I'm getting the feeling this might have something to do with a use case/problem I was having. Making the pod_subnet_chunk_size a parameter is great for small deployments, such as when the total pod_subnet is only a /24.

Maybe one additional question: since the first IP address of de pod_subnet can't be used ( it is reserved for the pervasive gateway of the bridge domain ), and possibly the last IP address neither, what happens with the last part of e.g. a /24 as a pod_subnet size?

To elaborate, using a pod_subnet of 192.168.0.0/24 and a pod_subnet_chunk_size of lets say 64, this would give the following ranges:
192.168.0.1 - pervasive gateway
192.168.0.2 - 192.168.0.65 -> first node
192.168.0.66 - 192.168.0.129 -> second node
192.168.0.130 - 192.168.0.193 ->third node
192.168.0.193 - ??? -> what happens with this? It is too small for 32 additonal IP's. Is this space lost?

Kind Regards,
Frederik Suijs.

PR #191, opened 2018-12-04
Allow user to specify alias in the ACI tenant

kahou82:
This PR is to allow user to specify alias in the ACI tenant


snaiksat:
Per Kahou, this is no longer needed, hence closing it.

PR #196, opened 2018-12-10
Fix handling of podnet annotation update, by using consistent data

jojimt:
Podnet annotation update causes IPAM cache to be messed up because there is in-flight data. Fix it by using local data under lock protection.

### DIFF
--- a/pkg/hostagent/ipam.go
+++ b/pkg/hostagent/ipam.go
@@ -18,7 +18,6 @@ package hostagent
 
 import (
 	"encoding/json"
-	"errors"
 	"fmt"
 	"net"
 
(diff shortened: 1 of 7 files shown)


pkg/hostagent/ipam.go
```diff
@@ -37,18 +36,35 @@ func combine(ranges []*ipam.IpAlloc) *ipam.IpAlloc {
 	return result
 }
 
-// must have index lock
-func (agent *HostAgent) rebuildIpam() {
+// builds the used IP info from metadata, at init.
+func (agent *HostAgent) buildUsedIPs() {
```
snaiksat: No locking is needed to protect the map used in this method?

jojimt: This is only executed at start up to build the map from the ep files. At that time, the agent is not yet running. Thus, lock is not necessary.

pkg/hostagent/ipam.go
```diff
@@ -37,18 +36,35 @@ func combine(ranges []*ipam.IpAlloc) *ipam.IpAlloc {
...
 			}
 		}
 	}
 
+	agent.log.Infof("buildIPUsed: %v addresses found", len(agent.usedIPs))
+}
+
+// must have index lock
+func (agent *HostAgent) rebuildIpam() {
+	for ip := range agent.usedIPs {
```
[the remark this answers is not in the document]
jojimt: Earlier, epMetadata was the only record of the IP allocation and that was the problem as it does not get updated synchronously. That is now replaced with the usedIPs map which is kept in sync with the IP allocation under lock. The epMeta is only used at start up for retrieving persistent information now.

PR #197, opened 2018-12-11
Modified acc_provision.py to consume a yaml flavors file

apoorva11029:
Added flavors.yaml instead of the flavors.py


provision/acc_provision/acc_provision.py
```diff
@@ -36,13 +33,16 @@
...
 ]
 
-VERSIONS = flavors.VERSIONS
-
-DEFAULT_FLAVOR_OPTIONS = flavors.KubeFlavorOptions
-
-CfFlavorOptions = flavors.CfFlavorOptions
-
-FLAVORS = flavors.FLAVORS
+with open("acc_provision/flavors.yaml", 'r') as stream:
```
snaiksat: Can we add an argument to acc-provision which will take a URL for this file? If no argument is provided, the default will be the file at this relative path.

snaiksat:
@puppet-py the motivation is to be able to, in the future, maintain this file separately. (that way it's lifecycle is independent of the acc-provision code)

PR #198, opened 2018-12-13
Adding various artifacts to k8s cluster-report

gaurav-dalvi:
Resolves: https://github.com/noironetworks/support/issues/789

### DIFF
--- a/cmd/acikubectl/cmd/debug.go
+++ b/cmd/acikubectl/cmd/debug.go
@@ -23,7 +23,9 @@ import (
 	"io"
 	"os"
 	"os/exec"
+	"strings"
 	"time"
+	"sync"
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/kubernetes"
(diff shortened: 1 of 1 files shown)


cmd/acikubectl/cmd/debug.go
```diff
@@ -190,36 +227,37 @@ func clusterReport(cmd *cobra.Command, args []string) {
...
-			continue
-		}
+	// Execute kubectl commands
+	for _, cmd := range cmds {
+		go func(cmd reportCmdElem, tarWriter *tar.Writer){
+			buffer := new(bytes.Buffer)
+			now := time.Now()
+			defer wg.Done()
+
+			fmt.Fprintln(os.Stderr, "Running command: kubectl", strings.Join(cmd.args, " "))
```
puppet-py: Request: <srings> pkg seems to be getting used only in this "one line (Join)" in the entire file.  Can we avoid that?
Possibly if we construct the args with single element then this join can be avoided, somewhat like this:
args: []string{"describe nodes -n " + systemNamespace}
args: []string{"describe daemonset aci-containers-host -n " + systemNamespace}

puppet-py: I meant the "strings" pkg

cmd/acikubectl/cmd/debug.go
```diff
@@ -87,6 +89,41 @@ func clusterReport(cmd *cobra.Command, args []string) {
 			name: "cluster-report/logs/controller/acc.log",
 			args: accLogCmdArgs(systemNamespace),
 		},
+		{
+			name: "cluster-report/status/describe_nodes.log",
```
puppet-py: Request: if can keep consistency in the name of the files. For "all describe resources" of k8s we are creating <resource>_status.log, can we do the same for the "describe nodes" ?

cmd/acikubectl/cmd/debug.go
```diff
@@ -190,36 +227,37 @@ func clusterReport(cmd *cobra.Command, args []string) {
...
+	for _, cmd := range cmds {
+		go func(cmd reportCmdElem, tarWriter *tar.Writer){
+			buffer := new(bytes.Buffer)
+			now := time.Now()
+			defer wg.Done()
+
+			fmt.Fprintln(os.Stderr, "Running command: kubectl", strings.Join(cmd.args, " "))
+			err := execKubectl(cmd.args, buffer)
+			if err != nil {
+				fmt.Fprintln(os.Stderr, err)
```
snaiksat: Don't you need to bail out here?

gaurav-dalvi: Thanks for catching. Done.

PR #199, opened 2018-12-13
Option to pass flavors.yaml as a URL or a filepath 

apoorva11029:
Adding option to pass flavors.yaml (for acc_provision.py) as a URL or a filepath 

The path/URL can be added to the user config input file as:
flavors_url:
   path: respective_path/url

The filepath/URL is handled in this sequence:
1. Check if it's a valid URL with valid yaml format. Simple check added: if the yaml has the 'default_flavor' field -- suggestions welcome
2. If 1. fails, do a similar check assuming flavors_url is a local filepath.
3. If both fail, default to the acc_provision/flavors.yaml saved in our repo.

Also, added a testdata/flavors_url.inp.yaml with an illegal URL address as flavors_url. 
In future, when we do have a hosted yaml online, can add a test with the correct URL too.

### DIFF
--- a/provision/acc_provision/acc_provision.py
+++ b/provision/acc_provision/acc_provision.py
@@ -1,5 +1,4 @@
 #!/usr/bin/env python
-
 from __future__ import print_function, unicode_literals
 
 import argparse
@@ -7,6 +6,7 @@ import base64
 import copy
 import functools
 import ipaddress
+import requests
 import json
 import os
 import os.path
@@ -18,31 +18,12 @@ import uuid
 import pkg_resources
 import pkgutil
 import yaml
+from yaml import SafeLoader
 
 from OpenSSL import crypto
 from .apic_provision import Apic, ApicKubeConfig
 from jinja2 import Environment, PackageLoader
 from os.path import exists
-
-
-VERSION_FIELDS = [
-    "cnideploy_version",
-    "aci_containers_host_version",
-    "opflex_agent_version",
-    "aci_containers_controller_version",
-    "openvswitch_version",
-]
-
-with open("acc_provision/flavors.yaml", 'r') as stream:
-        try:
-                doc = yaml.load(stream)
-        except yaml.YAMLError as exc:
-                print(exc)
-        DEFAULT_FLAVOR = doc['default_flavor']
-        VERSIONS = doc['versions']
-        DEFAULT_FLAVOR_OPTIONS = doc['kubeFlavorOptions']
-        CfFlavorOptions = doc['cfFlavorOptions']
-        FLAVORS = doc['flavors']
 
 
 def info(msg):
@@ -155,8 +136,21 @@ def config_default():
             "hostagent_log_level": "info",
             "opflexagent_log_level": "info",
         },
+        "flavors_url": {
+            "path": "",
+        },
     }
     return default_config
+
+
+# Adding a constructor to yaml SafeLoader so that it always overrides
+# the PyYAML handling of strings. For eg, strings like '10.0.0.0/24'
+# in our user input YAML files which contain masks. 
+def construct_yaml_str(self, node):
+    return self.construct_scalar(node)
+
+
+SafeLoader.add_constructor(u'tag:yaml.org,2002:str', construct_yaml_str)
 
 
 def config_user(config_file):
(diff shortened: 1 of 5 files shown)


provision/acc_provision/acc_provision.py
```diff
@@ -950,11 +931,88 @@ def provision(args, apic_file, no_random):
...
+        # try as a URL
+        res = requests.get(flavors_url)
+        if res and res.status_code == 200:
+            valid_yaml = check_yaml(res.text)
+            if valid_yaml:
+                info("Loading flavors from URL: " + flavors_url)
+                set_flavors(valid_yaml)
+                flag = True
+        else:
+            raise Exception
```
gaurav-dalvi: can we raise specific Exception ? Catching generic exception is not good.

apoorva11029: Should I raise one of the built-in exceptions? Or would defining custom be better because I need to catch both bad URL and bad YAML errors

provision/acc_provision/acc_provision.py
```diff
@@ -950,11 +931,88 @@ def provision(args, apic_file, no_random):
...
+                    flag = True
+    finally:
+        # try the static file in repo
+        if not flag:
+            return default_flavors_path()
+
+
+def set_flavors(doc):
+    global DEFAULT_FLAVOR, VERSIONS, DEFAULT_FLAVOR_OPTIONS, CfFlavorOptions, FLAVORS
+    DEFAULT_FLAVOR = doc['default_flavor']
```
snaiksat: I know its not directly to this commit, but could you also update the following:
https://github.com/noironetworks/aci-containers/blob/master/provision/acc_provision/flavors.yaml#L1
to: kubernetes-1.12

Please also update the commit message accordingly.

snaiksat:
BTW, in the commit message did you mean to say:
relative_path
where you current have:
respective_path

provision/acc_provision/acc_provision.py
```diff
@@ -950,11 +931,88 @@ def provision(args, apic_file, no_random):
...
 def main(args=None, apic_file=None, no_random=False):
     # apic_file and no_random are used by the test functions
     if args is None:
         args = parse_args()
 
+    config = config_user(args.config)
+    if config:
+        if 'flavors_url' not in config:
+            config['flavors_url'] = {}
+            config['flavors_url']['path'] = "blah"
```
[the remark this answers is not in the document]
apoorva11029: Assigned a specific(absurd) value to check for it later in the code, so that we know the code set it. 
In this case I guess it shouldn't matter because we anyway want to default to the flavors.yaml in our path. I could change it to None as well.

apoorva11029:
@snaiksat I should have used the required_path or just path. My bad. It doesn't have to the relative path though.

provision/acc_provision/acc_provision.py
```diff
@@ -950,11 +931,88 @@ def provision(args, apic_file, no_random):
...
+    "aci_containers_host_version",
+    "opflex_agent_version",
+    "aci_containers_controller_version",
+    "openvswitch_version",
+]
+
+
+def check_yaml(text_to_verify):
+    try:
+        flavors_yaml = yaml.load(text_to_verify)
```
gaurav-dalvi: safe_load is good here. bandit tool flags this :

```
>> Issue: [B506:yaml_load] Use of unsafe yaml load. Allows instantiation of arbitrary objects. Consider yaml.safe_load().
   Severity: Medium   Confidence: High
   Location: ./provision/acc_provision/acc_provision.py:38
   More Info: https://bandit.readthedocs.io/en/latest/plugins/b506_yaml_load.html
37	        try:
38	                doc = yaml.load(stream)
39	        except yaml.YAMLError as exc:

--------------------------------------------------
>> Issue: [B506:yaml_load] Use of unsafe yaml load. Allows instantiation of arbitrary objects. Consider yaml.safe_load().
   Severity: Medium   Confidence: High
   Location: ./provision/acc_provision/acc_provision.py:167
```

provision/acc_provision/acc_provision.py
apoorva11029: Included in the latest commit.

provision/acc_provision/acc_provision.py
apoorva11029: Modified and added in commit message

puppet-py:
The python2.7 failure wrt AddressValueError Exception, try using Unicode(u'x.x.x.x/mask'). Have encountered this error myself :-)

apoorva11029:
Making a new PR from my fork with the changes requested by peer reviews.

PR #204, opened 2018-12-14
Include flavors.yaml to the python package

apoorva11029:
Modified provision/MANIFEST.in to include all yaml files under acc_provision


puppet-py:
+1

PR #206, opened 2018-12-14
Removing cluster-info dump command from acikubectl cluster-report

gaurav-dalvi:


gaurav-dalvi:
tests:
1: Built acikubectl using make dist-static/acikubectl
2: Tested on vk8s_1 and it works fine.

PR #210, opened 2018-12-16
Removing goroutine call in acikubectl cluster-report command

gaurav-dalvi:


snaiksat:
@jojimt we are seeing a couple of issues with the previous commit, (1) commands were timing out when doing everything in parallel, (2) the implementation it resulted in all or nothing (failure was not graceful). This PR is reverting the implementation to the way it was prior to the change to making everything execute in parallel, and was at least known to be working in the field. We can separately explore as to what level of parallelism, and/or implementation logic will work in a more robust way.

jojimt:
Ah ok. Didn’t realize it was a revert.

PR #216, opened 2019-01-01
Build aci-containers and opflex private images in non jenkins environment.

mchalla:
Currently the only way to build these images is via Jenkins that is time consuming since you need to clone a pipeline without messing with the repo images. Secondly you cannot access the offical private repo without vpn to Cisco. This script simplifies private build.


build-priv.sh
```diff
@@ -0,0 +1,72 @@
...
+# 2. git clone https://github.com/opendaylight/opflex
+# 3. Modify docker/Dockerfile-* and Makefile to reflect DOCKER_USER
+# 4. docker login as DOCKER_USER
+# usage: build.sh <opflex-dir> <aci-containers-dir> <docker-user>
+# example: ./build-priv.sh /home/mchalla/work/opflex-noiro /home/mchalla/go/src/github.com/noironetworks/aci-containers challa
+
+OPFLEX_DIR=$1
+ACICONTAINERS_DIR=$2
+DOCKER_USER=$3
+
```
jojimt: Instead of forcing the command line order this way, can we just use the variable directly with some defaults?

e.g. 
: ${OPFLEX_DIR:=${HOME}/work/opflex-noiro}
: ${ACICONTAINERS_DIR:=${GOPATH}/src/github.com/noironetworks/aci-containers}

It will simplify the usage.

mchalla: Yep. Changed it to use defaults if not set.

build-priv.sh
```diff
@@ -0,0 +1,78 @@
+#!/bin/bash
+# Build opflex and aci-containers private images in non jenkins environment
+# 1. go get github.com/noironetworks/aci-containers
+# 2. mkdir -p $HOME/work && cd $HOME/work
+# 3. git clone https://github.com/noironetworks/opflex opflex-noiro
+# 4. Modify docker/Dockerfile-* and Makefile to reflect DOCKER_USER
+# 5. docker login as DOCKER_USER
+# usage: build.sh <opflex-dir> <aci-containers-dir> <docker-user>
+# example: ./build-priv.sh challa
+
```
gaurav-dalvi: Good to add `set -euxo pipefail`
https://vaneyckt.io/posts/safer_bash_scripts_with_set_euxo_pipefail/

build-priv.sh
```diff
@@ -0,0 +1,78 @@
...
+DOCKER_USER=$1
+
+[ -z "$GOPATH" ] && GOPATH=$HOME/go
+export GOPATH
+ACICONTAINERS_DIR=$GOPATH/src/github.com/noironetworks/aci-containers
+
+[ -z "$OPFLEX_DIR" ] && OPFLEX_DIR=$HOME/work/opflex-noiro
+export OPFLEX_DIR
+
+set -x
```
gaurav-dalvi: here as well. `set -euxo pipefail`

docker/Dockerfile-opflexserver
```diff
@@ -0,0 +1,9 @@
+FROM alpine:3.7
```
gaurav-dalvi: Good to use exact version rather than relying on `latest` tag. Thanks.

docker/launch-opflexserver.sh
```diff
@@ -0,0 +1,21 @@
+#!/bin/sh
+
+set -e
+set -x
```
gaurav-dalvi: same here.

build-priv.sh
mchalla: Thank you for the link. Added that.

PR #218, opened 2019-01-02
Adding gometalinter checks

gaurav-dalvi:
How to run:
1: `make gometalintercheck`
2: `./golang_checks.sh`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -97,7 +97,7 @@ goinstall:
 	${INSTALL_CMD} ${BASE}/cmd/hostagent
 	${INSTALL_CMD} ${BASE}/cmd/acikubectl
 
-dist/opflex-agent-cni: ${AGENTCNI_DEPS} 
+dist/opflex-agent-cni: ${AGENTCNI_DEPS}
 	${BUILD_CMD} -o $@ ${BASE}/cmd/opflexagentcni
 dist-static/opflex-agent-cni: ${AGENTCNI_DEPS}
 	${STATIC_BUILD_CMD} -o $@ ${BASE}/cmd/opflexagentcni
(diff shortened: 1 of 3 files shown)


golang_checks.sh
```diff
@@ -0,0 +1,48 @@
...
+
+# A deadline of 10m may seem excessive, but on slow machines it's preferable to allow it to take longer than to fail
+# the build.
+gometalinter --deadline=600s --config=./gometalinter.json --vendor ./... > metalinter-required.log 2>&1 &
+METALINTER_REQUIRED_PID="$!"
+
+gometalinter --deadline=90s --vendor --disable-all --enable=goconst --enable=errcheck --enable=dupl \
+    --enable=vet --enable=goimports ./... > metalinter-optional.log 2>&1 &
+METALINTER_OPTIONAL_PID="$!"
+
```
jojimt: Why are we using two separate invocations? Can't we run everything one pass?

Makefile
```diff
@@ -97,7 +97,7 @@ goinstall:
 	${INSTALL_CMD} ${BASE}/cmd/hostagent
 	${INSTALL_CMD} ${BASE}/cmd/acikubectl
 
-dist/opflex-agent-cni: ${AGENTCNI_DEPS} 
+dist/opflex-agent-cni: ${AGENTCNI_DEPS}
```
jojimt: Not related to this PR. But, we should get rid of the deps business I think. It is not consistent with go. We should use deps only for the secondary artifacts generated from go binaries.

golang_checks.sh
gaurav-dalvi: Good point. I will remove it in next version

PR #224, opened 2019-01-14
Add functionality to get git release info using acikubectl

apoorva11029:
To get the git commit ID and build time, run on any host:
acikubectl version
or
acikubectl version --kubeconfig=required_path

Git commit info is saved inside the aci-containers-controller container  and can be accessed by running "aci-containers-controller --version"

### DIFF
--- a/cmd/acikubectl/cmd/version.go
+++ b/cmd/acikubectl/cmd/version.go
@@ -1,0 +1,66 @@
+// Copyright © 2019 Cisco Systems, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or
+// implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package cmd
+
+import (
+	"bytes"
+	"fmt"
+	"github.com/spf13/cobra"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"os"
+	"strings"
+)
+
+func getVersion(cmd *cobra.Command, args []string) {
+	if len(args) != 0 {
+		fmt.Fprintln(os.Stderr, "More arguments than required")
+		return
+	}
+	kubeClient := initClientPrintError()
+	if kubeClient == nil {
+		return
+	}
+	systemNamespace, err := findSystemNamespace(kubeClient)
+	if err != nil {
+		fmt.Fprintln(os.Stderr, "Could not find aci-containers system namespace:", err)
+		return
+	}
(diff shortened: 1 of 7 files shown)


cmd/acikubectl/cmd/version.go
```diff
@@ -0,0 +1,67 @@
+// Copyright © 2017 Cisco Systems, Inc.
```
jojimt: 2019 :)

cmd/acikubectl/cmd/version.go
```diff
@@ -0,0 +1,67 @@
...
+        for _, pod := range systemNamespacePods.Items {
+                if strings.Contains(pod.Name, "aci-containers-controller") {
+                    buffer := new(bytes.Buffer)
+                    mylist := []string{"exec", "-n" + systemNamespace, pod.Name, "--", "/bin/sh", "-c", "get-gitCommit"}
+                    execKubectl(mylist, buffer)
+                    fmt.Println(buffer.String())
+                    break
+                }
+        }
+        return
```
jojimt: Please eliminate the redundant return.

pkg/controller/controller.go
```diff
@@ -274,6 +275,19 @@ func (cont *AciController) Run(stopCh <-chan struct{}) {
 	var err error
 	var privKey []byte
 	var apicCert []byte
+
```
jojimt: Please move this to cmd/controller/main.go. If you add --version as a cli flag there, you can also eliminate the shell script. Instead, you could just invoke controller --version which would just print the version and exit.

scripts/getGitCommit.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+
+set -euo pipefail
+
+if command -v git &>/dev/null && git rev-parse &>/dev/null; then
+    GIT_COMMIT=$(git rev-parse HEAD)
```
jojimt: Can you add these lines as well? 
        if [ -n "$(git status --porcelain --untracked-files=no)" ]; then
                GIT_COMMIT="$GIT_COMMIT-unsupported"
        fi

cmd/acikubectl/cmd/version.go
jojimt: Also, please run gofmt -s -w on all modified go files.

cmd/acikubectl/cmd/version.go
apoorva11029: Done, thanks for reminding!

cmd/acikubectl/cmd/version.go
apoorva11029: Added in the latest commit

pkg/controller/controller.go
apoorva11029: Thanks for the suggestion, made life easier. Removed the script and added the --version cli flag to aci-containers-controller

scripts/getGitCommit.sh
apoorva11029: Not adding it here because our current build leaves some untracked files

cmd/acikubectl/cmd/version.go
```diff
@@ -0,0 +1,66 @@
+// Copyright © 209 Cisco Systems, Inc.
+//
```
jojimt: Typo -- should be 2019

apoorva11029: Yes thanks, fixed.

PR #225, opened 2019-01-14
Moving version information from flavors.yaml to versions.yaml

gaurav-dalvi:
Resolving this : https://github.com/noironetworks/support/issues/810

### DIFF
--- a/provision/acc_provision/acc_provision.py
+++ b/provision/acc_provision/acc_provision.py
@@ -27,6 +27,16 @@ from jinja2 import Environment, PackageLoader
 from os.path import exists
 
 
+# This black magic forces pyyaml to load YAML strings as unicode rather
+# than byte strings in Python 2, thus ensuring that the type of strings
+# is consistent across versions.  From
+# https://stackoverflow.com/a/2967461/3857947.
+def construct_yaml_str(self, node):
+    return self.construct_scalar(node)
+
+
+SafeLoader.add_constructor(u'tag:yaml.org,2002:str', construct_yaml_str)
+
 VERSION_FIELDS = [
     "cnideploy_version",
     "aci_containers_host_version",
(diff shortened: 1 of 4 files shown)


provision/acc_provision/acc_provision.py
```diff
@@ -37,15 +38,22 @@
...
 
 FLAVORS_PATH = os.path.dirname(os.path.realpath(__file__)) + "/flavors.yaml"
+VERSIONS_PATH = os.path.dirname(os.path.realpath(__file__)) + "/versions.yaml"
+
 
+with open(VERSIONS_PATH, 'r') as stream:
+    try:
+        doc = yaml.safe_load(stream)
+    except yaml.YAMLError as exc:
+        print(exc)
```
puppet-py: standard logger func for error is "err". We should use "err" instead of print.
suggesting the same for line 55

utils/python/edit_versions_yaml.py
```diff
@@ -0,0 +1,53 @@
...
+                      help="last successful build for OpFlex-container")
+    parser.add_option("--ovs-container",
+                      dest="ovs_container",
+                      help="last successful build for OVS-container")
+
+    (opts, _) = parser.parse_args()
+    if (not opts.release_branch or not opts.acicontainers_tools or not opts.opflex_container or not opts.ovs_container):
+        parser.error("Required parameter(s) missing")
+
+    orig_data = read_yaml_file(VERSIONS_PATH)
```
puppet-py: should we not have a null-check for VERSIONS_PATH ? Incase the relative path is wrong or versions.yaml is missing. This can be handled in a unit-test though, since I did not see the unit-test hence raising it

provision/acc_provision/acc_provision.py
gaurav-dalvi: Good catch.

utils/python/edit_versions_yaml.py
gaurav-dalvi: Context manager does that for us. If `VERSIONS_PATH` file is not present then weget following type of error.
```
python
Python 2.7.10 (default, Oct  6 2017, 22:29:07)
[GCC 4.2.1 Compatible Apple LLVM 9.0.0 (clang-900.0.31)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> with open('foo', 'r') as file:
...     data = file.readlines()
...
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
IOError: [Errno 2] No such file or directory: 'foo'
```

PR #234, opened 2019-01-29
Removing following options from acikubectl debug command

gaurav-dalvi:
1: logs
2: node-cmd
Resolving : CSCvo09390


snaiksat:
Thanks for the changes. Could you please update the commit message to indicate specifically which logs were previously fetched by the commands that we are removing, and how those logs can be fetched using the functionality we have added in the recent commits? For instance, are we capturing the output of all the OVS commands that were being run by the commands we are removing here?

snaiksat:
Great, thanks for making the commit changes!

PR #235, opened 2019-01-29
Added missing RBAC policies for host-agent

puppet-py:
Found during OCP-3.11 install
- aci-containers-host-agent could not access the following
  resources: namespace, networkpolicies and deployments
- Added RBAC policies for the above resources


gaurav-dalvi:
@jojimt : Done. Could you please take a look ?

PR #245, opened 2019-02-04
Moving to OVS 2.10.1

snaiksat:
Using alpine 3.9 as base image

### DIFF
--- a/docker/Dockerfile-openvswitch
+++ b/docker/Dockerfile-openvswitch
@@ -1,4 +1,4 @@
-FROM alpine:3.8
+FROM alpine:3.9
 RUN apk upgrade --no-cache && \
   apk --no-cache add openvswitch logrotate
 COPY docker/launch-ovs.sh docker/liveness-ovs.sh dist-static/ovsresync /usr/local/bin/


mchalla:
I understand that ovs is backward and forward compatible when it comes to using any version of the kernel module with any version of ovs. I am not completely sure the same applies to ovs commands running against a certain version of ovs. (The recommendation is to recompile commands with that version) Since agent uses ovs headers similar to the commands I believe a similar concern exists for the agent. I am wondering what kind of testing we should be doing to qualify the upgrades since recompiling our code with newer headers is not straightforward. Thoughts ?

mchalla:
Basically when we change agent, make check compares known set of flows against the changed set and fails if anything changes. We need something similar here, although we cannot unfortunately compile agent with newer headers. May be an alternative is to run a decent feature set with older version of ovs, get the flow dump and compare the same feature set against the newer version of ovs. Other than that its fine with me. I also feel we should look into upgrading agent to use the newer headers.

snaiksat:
I believe the past approach has been to fix things if found broken. I understand that a similar approach was followed for moving to using OVS 2.9 for OpenStack. Adding @mandeepdhami to this review since the move to 2.10 was suggested by him.

mchalla:
I guess it should be fine to merge it then.

PR #249, opened 2019-02-07
Add default subscription refresh time of 80 seconds

ebati:
Hi,

`ApicRefreshTimer` documentation states `"0"` is a valid value but its value is used while creating `time.Ticker` which leads to panic. I added default value of 80 seconds as noted in comment [here](https://github.com/noironetworks/aci-containers/blob/32670167b8129752eec0a06f6f15334c2fc4e026/pkg/controller/config.go#L71) 

I also find (same comment)

 > NOTE: the subscription timeout is not supported by APIC versions before 3.2(3), so this value must not be set

misleading as this value should be set to `"0"` instead of leaving unset (leaving unset sends refresh time parameter with 900secs which results in `Bad Request`). I didn't changed this comment as English is not my primary language and i might be wrong.

### DIFF
--- a/pkg/apicapi/apicapi.go
+++ b/pkg/apicapi/apicapi.go
@@ -38,6 +38,10 @@ import (
 	"github.com/juju/ratelimit"
 )
 
+// defaultConnectionRefresh is used as connection refresh interval if
+// RefreshInterval is set to 0
+const defaultConnectionRefresh = 30 * time.Second
+
 func complete(resp *http.Response) {
 	io.Copy(ioutil.Discard, resp.Body)
 	resp.Body.Close()
@@ -423,7 +427,11 @@ func (conn *ApicConnection) runConn(stopCh <-chan struct{}) {
 		}()
 	}
 
-	refreshTicker := time.NewTicker(conn.RefreshInterval)
+	refreshInterval := conn.RefreshInterval
+	if refreshInterval == 0 {
+		refreshInterval = defaultConnectionRefresh
+	}
+	refreshTicker := time.NewTicker(refreshInterval)
 	defer refreshTicker.Stop()
 
 	closeConn := func(stop bool) {
@@ -811,11 +819,11 @@ func (conn *ApicConnection) subscribe(value string, sub *subscription) bool {
 		kind = "class"
 	}
 
-        refresh_interval := ""
-        if conn.RefreshInterval != 0 {
-                refresh_interval = fmt.Sprintf("refresh-timeout=%s&",
-                        conn.RefreshInterval)
-        }
+	refresh_interval := ""
+	if conn.RefreshInterval != 0 {
+		refresh_interval = fmt.Sprintf("refresh-timeout=%s&",
+			conn.RefreshInterval)
+	}
 
 	// properly encoding the URI query parameters breaks APIC
 	uri := fmt.Sprintf("/api/%s/%s.json?subscription=yes&%s%s",


snaiksat:
@ebati Thanks for your PR. Could you please send me to the panic trace that you are seeing when you set the option to 0 (i.e. zero)?

ebati:
Related option from yaml:
  - aci-containers-controller `image: noiro/aci-containers-controller:1.9r32`
  - controller-config has `"apic-refreshtime": "0",` set

log:
```
...
panic: non-positive interval for NewTicker

goroutine 40 [running]:
time.NewTicker(0x0, 0x14f4ab0)
	/usr/local/go/src/time/tick.go:23 +0x1a1
github.com/noironetworks/aci-containers/pkg/apicapi.(*ApicConnection).runConn(0xc42032be60, 0xc42009e0c0)
	/opt/jenkins/workspace/team_noiro_engineering/AciContainers-container-kube19/src/github.com/noironetworks/aci-containers/pkg/apicapi/apicapi.go:426 +0x487
github.com/noironetworks/aci-containers/pkg/apicapi.(*ApicConnection).Run.func1(0xc42032be60, 0xc42009e0c0)
	/opt/jenkins/workspace/team_noiro_engineering/AciContainers-container-kube19/src/github.com/noironetworks/aci-containers/pkg/apicapi/apicapi.go:516 +0x628
github.com/noironetworks/aci-containers/pkg/apicapi.(*ApicConnection).Run(0xc42032be60, 0xc42009e0c0)
	/opt/jenkins/workspace/team_noiro_engineering/AciContainers-container-kube19/src/github.com/noironetworks/aci-containers/pkg/apicapi/apicapi.go:517 +0x44
created by github.com/noironetworks/aci-containers/pkg/controller.(*AciController).Run
	/opt/jenkins/workspace/team_noiro_engineering/AciContainers-container-kube19/src/github.com/noironetworks/aci-containers/pkg/controller/controller.go:407 +0xc84
```

snaiksat:
Thanks @ebati. Looking into it. The initialization not happening correctly when set to 0 is definitely a bug.

snaiksat:
@ebati We have split acc-provision into its own repo. Would you like to move your PR to that repo?
https://github.com/noironetworks/acc-provision

pkg/apicapi/apicapi.go
```diff
@@ -423,7 +425,12 @@ func (conn *ApicConnection) runConn(stopCh <-chan struct{}) {
 		}()
 	}
 
-	refreshTicker := time.NewTicker(conn.RefreshInterval)
+	refreshTickTime := conn.RefreshInterval
+	if refreshTickTime == 0 {
```
snaiksat: Can we change this to:

if conn.RefreshInterval != 0 {
        refreshTicker := time.NewTicker(conn.RefreshInterval)
        defer refreshTicker.Stop()
}

pkg/apicapi/apicapi.go
```diff
@@ -811,11 +818,11 @@ func (conn *ApicConnection) subscribe(value string, sub *subscription) bool {
...
-        refresh_interval := ""
-        if conn.RefreshInterval != 0 {
-                refresh_interval = fmt.Sprintf("refresh-timeout=%s&",
-                        conn.RefreshInterval)
-        }
+	refresh_interval := ""
+	if conn.RefreshInterval != 0 {
+		refresh_interval = fmt.Sprintf("refresh-timeout=%s&",
+			conn.RefreshInterval)
+	}
```
snaiksat: I believe there is no change needed here, so this should not show up in the diff. Did something change with the formatting?

snaiksat:
@ebati Please ignore my previous comment about moving the PR to acc-provision, I was getting mixed up. Instead please see my subsequent inline comments on your PR.

pkg/apicapi/apicapi.go
ebati: It was spaces before but gofmt convert them to tabs.

ebati:
PTAL. I also rebased to master, i hope it is OK.

pkg/apicapi/apicapi.go
```diff
@@ -454,7 +459,7 @@ loop:
 		select {
 		case <-syncHook:
 			conn.fullSync()
-		case <-refreshTicker.C:
+		case <-refreshTickerChan:
```
snaiksat: Thanks for updating this part, I totally missed it. So thinking about this a little more, the refresh will never happen if conn.RefreshInterval is set to zero, right? I think we need to revert to using the original value when the config value is set to zero.

pkg/apicapi/apicapi.go
```diff
@@ -38,6 +38,10 @@ import (
 	"github.com/juju/ratelimit"
 )
 
+// defaultConnectionRefresh is used as connection refresh interval if
+// RefreshInterval is set to 0
+const defaultConnectionRefresh = 80 * time.Second
```
snaiksat: Apologies for the delay in reviewing this. The APIC side time out is 80 seconds, but we were earlier setting this to 30 seconds:
https://github.com/noironetworks/aci-containers/commit/fcb6e868f812b6234535454b0a10d5175c362c0d#diff-31ded243a50636001aa9734838222ca0L60

Since we want to refresh before the APIC timeout. Could you please change it to 30?

ebati: Updated to 30 seconds

snaiksat: Thanks @ebati!

PR #250, opened 2019-02-07
Removing custom description from setup.py for acc_provision module

gaurav-dalvi:
Its not recommended to put custom function calls to the description
https://docs.python.org/3/distutils/setupscript.html#additional-meta-data
as it goes on to pypi home page of that module.


snaiksat:
Merging this because its currently breaking the packaging.

PR #263, opened 2019-02-19
Add release info logs to host-agent container

apoorva11029:
git commit ID and build time info will be present in the container logs for host-agent and controller

### DIFF
--- a/cmd/controller/main.go
+++ b/cmd/controller/main.go
@@ -96,6 +96,13 @@ func main() {
 		panic(err.Error())
 	}
 
+	if controller.GetVersion().GitCommit != "" {
+		versionInfo := controller.GetVersion()
+		log.Info("Running controller built from git commit ID " +
+                          versionInfo.GitCommit + " at build time " +
+                          versionInfo.BuildTime)
+	}
+
 	cont := controller.NewController(config, env, log)
 	cont.Init()
 	cont.Run(wait.NeverStop)
--- a/cmd/hostagent/main.go
+++ b/cmd/hostagent/main.go
@@ -15,10 +15,13 @@
 package main
 
 import (
+	"bytes"
 	"encoding/json"
 	"errors"
 	"flag"
+	"fmt"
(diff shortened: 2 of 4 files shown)


cmd/controller/main.go
```diff
@@ -96,6 +96,13 @@ func main() {
 		panic(err.Error())
 	}
 
+	if controller.GetVersion().GitCommit != "" {
```
gaurav-dalvi: We are checking exact same thing here as well : https://github.com/noironetworks/aci-containers/blob/36fd98d6054b94c154c7facf1e5598707c167ec0/cmd/controller/main.go#L44 
Why do we need to repeat this ?

cmd/hostagent/main.go
```diff
@@ -80,6 +80,13 @@ func main() {
 		panic(err.Error())
 	}
 
+        if hostagent.GetVersion().GitCommit != "" {
+                versionInfo := hostagent.GetVersion()
+                log.Info("Running controller built from git commit ID " +
```
snaiksat: controller -> host-agent ?

pkg/controller/controller.go
```diff
@@ -275,10 +275,6 @@ func (cont *AciController) Run(stopCh <-chan struct{}) {
 	var privKey []byte
 	var apicCert []byte
 
-	if gitCommit != "" {
```
gaurav-dalvi: Is there any way in the logs developer can see the version / commit id after removing this ?

cmd/hostagent/main.go
apoorva11029: Right, fixed.

cmd/controller/main.go
apoorva11029: Need it for the version flag too

pkg/controller/controller.go
apoorva11029: The (equivalent)code has been moved to cmd/controller/main.go

PR #265, opened 2019-02-22
Delete epfile when cni device is deleted

jojimt:
Even after cni delete was performed (and the IP address was released), the epfile was not deleted until we received a pod delete event from the API server. This PR ensures that the epfile is deleted when the CNI interface is removed.

### DIFF
--- a/pkg/hostagent/pods.go
+++ b/pkg/hostagent/pods.go
@@ -270,6 +270,7 @@ func (agent *HostAgent) podChangedLocked(podobj interface{}) {
 		agent.epDeleted(&epUuid)
 		return
 	}
+	agent.cniToPodID[epMetaKey] = epUuid
 
 	epGroup, secGroup, _ := agent.assignGroups(pod)
 	epAttributes := pod.ObjectMeta.Labels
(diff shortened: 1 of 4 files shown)


pkg/hostagent/pods.go
```diff
@@ -386,3 +387,19 @@ func (agent *HostAgent) podDeletedLocked(obj interface{}) {
...
+	if !ok {
+		agent.log.Warnf("cniEpDelete: PodID not found for %s", cniKey)
+		return
+	}
+	delete(agent.cniToPodID, cniKey)
+
+	if _, ok := agent.opflexEps[epUuid]; ok {
+		delete(agent.opflexEps, epUuid)
+		agent.scheduleSyncEps()
+	}
```
snaiksat: May be a debug log here?

pkg/hostagent/pods.go
```diff
@@ -270,6 +270,7 @@ func (agent *HostAgent) podChangedLocked(podobj interface{}) {
 		agent.epDeleted(&epUuid)
 		return
 	}
+	agent.cniToPodID[epMetaKey] = epUuid
```
mchalla: Should we be mapping this against something more unique than pod namespace and pod name ? What if I restart the pod, get a new uuid but namespace and name are same ?

pkg/hostagent/pods.go
mchalla: Pl add a log here and the when we come via cni path and in the original code thats also attempting this delete.

pkg/hostagent/pods.go
jojimt: Namespace.name is unique within the cluster. Kubelet will not create a new CNI interface for a given Namespace.Name combination before deleting the old one.

pkg/hostagent/pods.go
```diff
@@ -386,3 +387,19 @@ func (agent *HostAgent) podDeletedLocked(obj interface{}) {
...
+	epUuid, ok := agent.cniToPodID[cniKey]
+	if !ok {
+		agent.log.Warnf("cniEpDelete: PodID not found for %s", cniKey)
+		return
+	}
+	delete(agent.cniToPodID, cniKey)
+
+	if _, ok := agent.opflexEps[epUuid]; ok {
+		delete(agent.opflexEps, epUuid)
+		agent.scheduleSyncEps()
```
mchalla: one more thing that looks like a design problem is why do we sync / walk all the eps every time on add or delete. The sync should happen either on a timer or when agent restarts. when we add or delete we exactly know which pod it is. Not sure why we need to walk that list.

pkg/hostagent/pods.go
mchalla: i see, thats fine then.

pkg/hostagent/pods.go
jojimt: Yes, this is per current design. The alternative is to generate deltas and send it to the sync thread. That can have races which will need to be explicitly addressed. That's probably why the original implementer chose the current approach.

PR #267, opened 2019-02-26
Add default subscription refresh time of 30 seconds

snaiksat:
(cherry picked from commit 81ddc1afebfacba33739782889df7be5a322ff03)

### DIFF
--- a/pkg/apicapi/apicapi.go
+++ b/pkg/apicapi/apicapi.go
@@ -38,6 +38,10 @@ import (
 	"github.com/juju/ratelimit"
 )
 
+// defaultConnectionRefresh is used as connection refresh interval if
+// RefreshInterval is set to 0
+const defaultConnectionRefresh = 30 * time.Second
+
 func complete(resp *http.Response) {
 	io.Copy(ioutil.Discard, resp.Body)
 	resp.Body.Close()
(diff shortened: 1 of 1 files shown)


pkg/apicapi/apicapi.go
```diff
@@ -38,6 +38,10 @@ import (
 	"github.com/juju/ratelimit"
 )
 
+// defaultConnectionRefresh is used as connection refresh interval if
+// RefreshInterval is set to 0
+const defaultConnectionRefresh = 30 * time.Second
```
gaurav-dalvi: Dont we need to change in the test code as well ?

o/p of grep
```
pkg/apicapi/apicapi_test.go
392:	conn.RefreshInterval = 5 * time.Millisecond

pkg/apicapi/apic_types.go
87:	RefreshInterval   time.Duration

pkg/apicapi/apicapi.go
196:		RefreshInterval:   time.Duration(refresh) * time.Second,
426:	refreshTicker := time.NewTicker(conn.RefreshInterval)
815:        if conn.RefreshInterval != 0 {
817:                        conn.RefreshInterval)
```

Also in which cases, conn.RefreshInterval := 0 ? Do we really need to check if condition at line 430 ?

snaiksat: It could be set to 0 in the confg.

snaiksat: Regarding changing UT, yes we could do that (in a separate PR since this is just a cherry-pick). However as far as the tests are concerned its just some value (which the test itself is setting), the tests are not really checking that its a particular value or attaching any meaning to that value.

PR #272, opened 2019-03-24
Option to enable security-sharing in ext subnets for k8s services

apoorva11029:
Default value: import-security

If a service contract is defined with the scope value "global", the respective extSubnet is registered with shared-security scope setting as well

This is an addition to https://github.com/noironetworks/aci-containers/pull/186 where we first added option to change service contract scope

Also modified the unit test to verify contract scope to include this code change. TestServiceAnnotation now also checks for the extSubnet's scope value when the service contract scope is set to global

### DIFF
--- a/pkg/controller/cf_update_handlers.go
+++ b/pkg/controller/cf_update_handlers.go
@@ -342,12 +342,9 @@ func (env *CfEnvironment) createAppServiceGraph(appId string, extIps []string) (
 		// 2. Service graph contract and external network
 		serviceObjs = append(serviceObjs,
 			apicExtNet(name, cont.config.AciVrfTenant,
-				cont.config.AciL3Out, extIps))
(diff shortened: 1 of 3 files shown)


pkg/controller/cf_update_handlers.go
```diff
@@ -342,12 +342,9 @@ func (env *CfEnvironment) createAppServiceGraph(appId string, extIps []string) (
 		// 2. Service graph contract and external network
 		serviceObjs = append(serviceObjs,
 			apicExtNet(name, cont.config.AciVrfTenant,
-				cont.config.AciL3Out, extIps))
+				cont.config.AciL3Out, extIps, DefaultServiceExtNetShared))
```
snaiksat: Minor comment: I think in a follow up patch we should ideally change the naming convention to be "ExtSubnet" instead of "ExtNet".

PR #276, opened 2019-04-11
Image fixes for UCP

Vikash082:


docker/launch-hostagent.sh
```diff
@@ -61,11 +61,10 @@ else
 fi
 ACC_MAC=$(get_mac veth_host)
 
+eth_interface=$(ip route get 8.8.8.8 | awk '{ print $5; exit }')
```
jojimt: Why is this change required? As long as we use the same VM, there should be no reason to change the interface. Is docker somehow using a different interface as the node IP?

In any case, we cannot use 8.8.8.8 to derive the route information as it won't work when proxies are used for outside access.

Vikash082: Ok, I am not sure how hard coded value is useful and serves all deployments. Please ignore the patch and let the script as it is.

PR #284, opened 2019-05-29
Fixing the format of refresh-timeout in APIC subscription request.

kishorejonnala:
Before:
URL: https://127.0.0.1:33211/api/mo/uni/tn-common.json?subscription=yes&refresh-time
out=1m0s&query-target=subtree&rsp-subtree=full&
After:
URL: https://127.0.0.1:33211/api/mo/uni/tn-common.json?subscription=yes&refresh-time
out=60&query-target=subtree&rsp-subtree=full&

### DIFF
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -297,6 +297,12 @@ func (cont *AciController) Run(stopCh <-chan struct{}) {
 	}
 	cont.log.Info("ApicRefreshTimer conf is set to: ", refreshTimeout)
 
+        // Bailout if the refreshTimeout is more than 12Hours
+        if refreshTimeout > (12*60*60) {
+                cont.log.Info("ApicRefreshTimer can't be more than 12Hrs")
+                panic(err)
+        }
+
 	// If not defined, default to 32
 	if cont.config.PodIpPoolChunkSize == 0 {
 		cont.config.PodIpPoolChunkSize = 32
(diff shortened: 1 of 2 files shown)


pkg/controller/controller.go
```diff
@@ -297,6 +297,12 @@ func (cont *AciController) Run(stopCh <-chan struct{}) {
 	}
 	cont.log.Info("ApicRefreshTimer conf is set to: ", refreshTimeout)
 
+        // Bailout if the refreshTimeout is more than 12Hours
```
jojimt: Can we validate this in the acc-provision? Instead of panic, we should just cap the refresh at (slightly less than) the max here.

kishorejonnala: @jojimt Actual refresh cap is 24Hrs from APIC side, but suggested limit is 12Hrs;
Are you saying instead of panic, just set it to 12*60*60 if its more than that and perhaps log a message indicating that?

kishorejonnala: Also, currently we don't have the way to config in acc-provision. I plan to add the support and yes, will validate on acc-provision side as well.

jojimt: > @jojimt Actual refresh cap is 24Hrs from APIC side, but suggested limit is 12Hrs;
> Are you saying instead of panic, just set it to 12_60_60 if its more than that and perhaps log a message indicating that?

Yes, exactly(just a one time msg).

kishorejonnala:
@snaiksat Yes, I will add UT as a follow-up patch. Let's merge this PR.
@jojimt The validation is added on acc-provision side (https://github.com/noironetworks/acc-provision/pull/64), the conclusion was not to implicitly change user intent, rather bail-out so that user is informed and can correct the config.

PR #286, opened 2019-06-10
Fixing a concurrency issue in handling websocket notify path.

kishorejonnala:
Issue:
While processing websocket notification(s) from APIC and if connection to APIC gets reset during the sametime,
ACC is deleting those to be processed Dns from APIC once the conntectivity is restored.

Fix:
Webscoket notifications are processed and "dn"s gets queued in conn.deltaQueue to be processed by another go-routine and this deltaQueue
is passed by value to that goRoutine. These Dns and the status of the object (update/delete)  are also stored in conn.pendingSubDnUpdate
structure to determine and re-push if some one modifies an object which is owned/managed by ACC. This pendingSubDnUpdate structure is
reset with each websocket connection restart. Since conn.deltaQueue is passed by value and that queue is not drained due to a blocking call
"getSubtreeDn" while processing that queue, rest of the Dns from that queue get's processed once the connectivity gets restored. By then,
pendingSubDnUpdate is reset due to which these Dns are inferred to be marked as deleted, hence ACC deletes these from APIC.

Fix is to not reset pendingSubDnUpdate when websocket connection is restarted, rather delete each entry gracefully before draiing it from deltaQueue.

Testing:
Given the number of variables involved (ws notification and apic connectivity loss has to happen back to back), its hard to reproduce with test-automation.
The issue has been reproduced manually on Cam's setup and fix has been verified on the same. Also, made sure that the fix has no impact on reconcile logic
by adding a new entry to one of the ACC managed contract's filter and it get's deleted/reconciled by ACC as expected.

### DIFF
--- a/pkg/apicapi/apicapi.go
+++ b/pkg/apicapi/apicapi.go
@@ -244,9 +244,16 @@ func (conn *ApicConnection) handleSocketUpdate(apicresp *ApicResponse) {
 						pendingKind = pendingChangeUpdate
 					}
 					conn.indexMutex.Lock()
+
+					conn.log.WithFields(logrus.Fields{
+						"dn":  obj.GetDn(),
+						"obj": obj,
+					}).Debug("Processing websocket notification for:")
+
 					conn.pendingSubDnUpdate[dn] = pendingChange{
 						kind:   pendingKind,
 						subIds: subIds,
+						isDirty: false,
 					}
 					if conn.deltaQueue != nil {
 						conn.deltaQueue.Add(dn)
@@ -295,7 +302,7 @@ func (conn *ApicConnection) handleQueuedDn(dn string) bool {
 	var requeue bool
 	conn.indexMutex.Lock()
 	pending, hasPendingChange := conn.pendingSubDnUpdate[dn]
-	delete(conn.pendingSubDnUpdate, dn)
+	conn.pendingSubDnUpdate[dn] = pendingChange{ isDirty: true }
 	obj, hasDesiredState := conn.desiredStateDn[dn]
 	conn.indexMutex.Unlock()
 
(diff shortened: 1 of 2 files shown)


pkg/apicapi/apicapi.go
```diff
@@ -264,6 +270,8 @@ func (conn *ApicConnection) restart() {
 		conn.log.Debug("Restarting connection")
 		close(conn.restartCh)
 		conn.restartCh = nil
+		conn.log.Debug("Resetting deltaQueue")
+		conn.deltaQueue = nil
```
snaiksat: Hate to be pedantic - In the commit message could you add some context/details regarding the issue we are noticing and how we expect resetting the queue to handle that?

pkg/apicapi/apicapi.go
```diff
@@ -305,6 +313,7 @@ func (conn *ApicConnection) handleQueuedDn(dn string) bool {
 		}
 	}
 
+	conn.log.Debug("<== respClasses::hasPendingChange::hasDesiredState==>:", respClasses, hasPendingChange, hasDesiredState)
```
snaiksat: nit: did you intend to add this log message to this commit? Asking because the log message string format is different from other log messages.

pkg/apicapi/apicapi.go
```diff
@@ -642,10 +654,10 @@ func (conn *ApicConnection) getSubtreeDn(dn string, respClasses []string,
...
 	}
 	for _, obj := range apicresp.Imdata {
-		//conn.log.WithFields(logrus.Fields{
-		//	"dn":  obj.GetDn(),
-		//	"obj": obj,
-		//}).Debug("Object updated on APIC")
+		conn.log.WithFields(logrus.Fields{
+			"dn":  obj.GetDn(),
+			"obj": obj,
+		}).Debug("Object updated on APIC")
```
snaiksat: I suspect this commented out because it was too verbose even at debug level. We might just need to watch out.

pkg/apicapi/apicapi.go
```diff
@@ -674,7 +686,7 @@ func (conn *ApicConnection) queueDn(dn string) {
 func (conn *ApicConnection) postDn(dn string, obj ApicObject) bool {
 	conn.log.WithFields(logrus.Fields{
 		"dn": dn,
-		//"obj": obj,
+		"obj": obj,
```
snaiksat: ditto

PR #292, opened 2019-06-26
Refresh WS subscriptions 5seconds earlier than actual refresh-timeout value

kishorejonnala:
Issue:
Current code refreshes WS subscriptions exactly on the last second of refresh-timeout. Depending on APIC load, its possible that these refresh messages get processed a little late and result in subscription refresh failure.

Fix:
To trigger subscription refreshes 5sec earlier than actual refresh-timeout value that is used as part of initial subscription.

The fix has been tested on a fab setup.

### DIFF
--- a/pkg/apicapi/apicapi.go
+++ b/pkg/apicapi/apicapi.go
@@ -165,7 +165,7 @@ func configureTls(cert []byte) (*tls.Config, error) {
 
 func New(log *logrus.Logger, apic []string, user string,
 	password string, privKey []byte, cert []byte,
-	prefix string, refresh int) (*ApicConnection, error) {
+	prefix string, refresh int, refreshTickerAdjust int) (*ApicConnection, error) {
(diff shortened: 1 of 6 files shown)


pkg/apicapi/apicapi.go
```diff
@@ -442,7 +442,13 @@ func (conn *ApicConnection) runConn(stopCh <-chan struct{}) {
 	if refreshInterval == 0 {
 		refreshInterval = defaultConnectionRefresh
 	}
-	refreshTicker := time.NewTicker(refreshInterval)
+	refreshTickerInterval := refreshInterval
+	// Adjust refreshTickerInterval by 5sec.
+	// To refresh the subscriptions 5sec early than actual refresh timeout value
+        if refreshInterval > (5 * time.Second) {
```
snaiksat: Might be preferable to make this a configurable and default to 5.

PR #294, opened 2019-07-03
Update opflex device subscription in controller

apoorva11029:
Remove the filters on class subscription in controller for opflex devices. Implement these filters inside the functions for the subscription hook (opflexDeviceChanged and opflexDeviceDeleted) instead.

This is to make sure we don't filter out updates from APIC for opflex device deletions, as these notifications come with the above filter values set as defaults.

This is fix to http://cdets.cisco.com/apps/dumpcr?identifier=CSCvo58994&content=summary&format=html

### DIFF
--- a/pkg/controller/services.go
+++ b/pkg/controller/services.go
@@ -693,77 +693,90 @@ func (cont *AciController) fabricPathLogger(node string,
 }
 
 func (cont *AciController) opflexDeviceChanged(obj apicapi.ApicObject) {
-	var nodeUpdates []string
 
-	cont.indexMutex.Lock()
-	nodefound := false
-	for node, devices := range cont.nodeOpflexDevice {
-		found := false
+	devType := obj.GetAttrStr("devType")
+	domName := obj.GetAttrStr("domName")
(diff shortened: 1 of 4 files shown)


pkg/controller/services.go
```diff
@@ -693,77 +693,90 @@ func (cont *AciController) fabricPathLogger(node string,
...
-	for node, devices := range cont.nodeOpflexDevice {
-		found := false
+	devType := obj.GetAttrStr("devType")
+	domName := obj.GetAttrStr("domName")
+	ctrlrName := obj.GetAttrStr("ctrlrName")
 
-		if node == obj.GetAttrStr("hostName") {
-			nodefound = true
-		}
+	if ((devType == cont.env.OpFlexDeviceType()) && (domName == cont.config.AciVmmDomain) && (ctrlrName == cont.config.AciVmmController)) {
```
snaiksat: nit: Would it help to have a debug statement here - something to the effect of "Processing Opflex device change notification"?

apoorva11029: Added, thanks!

PR #297, opened 2019-07-12
renamed the direcotry structure to aci.snat from aci.nodeinfo

jayaramsatya:


snaiksat:
Ideally we prefer this to not be SNAT specific, but since we are currently having issues making this work, we can temporarily move forward with this and with a plan to revisit.

PR #298, opened 2019-07-12
renamed the direcotry structure to aci.snat from aci.nodeinfo

jayaramsatya:


snaiksat:
Please remember to use -x option in cherry-pick so that the cherry-picked commit gets included in the commit message and its easy to track which commits have beenn back ported.

PR #299, opened 2019-07-14
SNAT changes for controller 

apoorva11029:
Create a svc graph for SNAT feature introduced in https://github.com/noironetworks/aci-containers/pull/296
Provision the snat graph the first time when a snatpolicy custom resource is applied. A typical snatpolicy looks like: 
apiVersion: aci.snat/v1
kind: SnatPolicy
metadata:
  name: examplesnatpolicy
spec:
  snatIp:
    - 10.3.174.174/32
    - 10.4.168.15
  selector:
     labels:
       -
         key: version
         value: development

"snatIp" field could be a list of one or more IP/subnets that will used for our SNAT implementation. The "labels" field lists the k8s resources that should be matched with this snatpolicy. 
For any additional snatpolicies that get applied, add to the existing snat graph's external subnet list. 
For further updates/deletes on snatpolicies, update the external subnet list accordingly.
If the last snatpolicy is deleted, clear the snat svc graph as well. 

By default, port range 5000-65000 is used on each node for SNAT in this implementation.


pkg/controller/services.go
```diff
@@ -535,6 +599,105 @@ func (cont *AciController) updateServiceDeviceInstance(key string,
 	return nil
 }
 
+func (cont *AciController) updateServiceDeviceInstanceSnat(key string) error {
```
snaiksat: There are at least parts of the function updateServiceDeviceInstance which are very similar to this one. We should attempt to refactor and avoid code duplication. This can be done as a follow up patch.

pkg/controller/services.go
```diff
@@ -392,6 +418,60 @@ func apicDevCtx(name string, tenantName string,
...
+	return filter
+}
+
+func apicFilterSnat(name string, tenantName string,
+        portSpec []portRangeSnat) apicapi.ApicObject {
+
+        filter := apicapi.NewVzFilter(tenantName, name)
+        filterDn := filter.GetDn()
+
+        for i, port := range portSpec {
```
kishorejonnala: No Need for a loop here

pkg/controller/services.go
```diff
@@ -392,6 +418,60 @@ func apicDevCtx(name string, tenantName string,
...
+
+                fe := apicapi.NewVzEntry(filterDn, strconv.Itoa(i))
```
kishorejonnala: Move it into a function and pass the protocol as param to avoid repeated code here?

pkg/controller/config.go
```diff
@@ -161,6 +161,8 @@ type ControllerConfig struct {
 	// a list of subnet/gateway CIDR addresses that cover the
 	// addresses in the node service IP pool
 	NodeServiceSubnets []string `json:"node-service-subnets,omitempty"`
+
```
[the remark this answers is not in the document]
apoorva11029: Added, thanks!

pkg/controller/environment.go
```diff
@@ -72,8 +77,12 @@ func NewK8sEnvironment(config *ControllerConfig, log *logrus.Logger) (*K8sEnviro
 	if err != nil {
 		return nil, err
 	}
-
-	return &K8sEnvironment{kubeClient: kubeClient}, nil
+	log.Debug("Initializing snat client")
+	snatClient, err := snatclientset.NewForConfig(restconfig)
+	if err !=nil {
```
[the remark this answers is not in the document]
apoorva11029: The initialization is logged already. If there's an error with snatClient creation, it will show up in the logs through https://github.com/noironetworks/aci-containers/blob/403d6b9e0791de3c55718d24172758bb292af9c8/cmd/controller/main.go#L95

pkg/controller/services.go
```diff
@@ -312,23 +312,42 @@ func apicExtNet(name string, tenantName string, l3Out string,
...
-				subnet.SetAttr("scope", sharedSecurityString)
+		ip, _, _ := net.ParseCIDR(ingress)
+		// If ingress is a subnet
+		if ip != nil {
+			if ip != nil && ip.To4() != nil {
+				subnet := apicapi.NewL3extSubnet(enDn, ingress)
+				if sharedSecurity {
+					subnet.SetAttr("scope", sharedSecurityString)
+				}
+				en.AddChild(subnet)
```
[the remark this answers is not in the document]
apoorva11029: Done, thanks!

pkg/controller/services.go
apoorva11029: Noted!

pkg/controller/controller.go
```diff
@@ -106,9 +110,11 @@ type AciController struct {
 	nodeOpflexDevice     map[string]apicapi.ApicSlice
 	nodePodNetCache      map[string]*nodePodNetMeta
 	serviceMetaCache     map[string]*serviceMeta
+	snatPolicyCache      map[string]*ContSnatPolicy
```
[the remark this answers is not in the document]
apoorva11029: Noted!

pkg/controller/services.go
apoorva11029: Fixed, thanks!

pkg/controller/services.go
apoorva11029: Done, thanks!

PR #301, opened 2019-08-05
Create snat ext service file on hosts

apoorva11029:
Generate a external service file called "snat-external.service "with MAC address and service IP of each host inside the services file on host-agent pod. Create the file on each host when the first snat policy is applied, and delete the file if there are no more snat policies applied.

### DIFF
--- a/pkg/hostagent/snats.go
+++ b/pkg/hostagent/snats.go
@@ -37,6 +37,10 @@ import (
 	"strings"
 )
 
+// Filename used to create external service file on host
+// example snat-external.service
+const SnatService = "snat-external"
+
 type OpflexPortRange struct {
 	Start int `json:"start,omitempty"`
 	End   int `json:"end,omitempty"`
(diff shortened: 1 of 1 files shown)


pkg/hostagent/snats.go
```diff
@@ -339,6 +352,50 @@ func (agent *HostAgent) snaGlobalInfoChanged(snatobj interface{}, logger *logrus
...
+			InterfaceVlan: uint16(agent.config.ServiceVlan),
+			ServiceMac: agent.serviceEp.Mac,
+			InterfaceIp: agent.serviceEp.Ipv4.String(),
+		}
+		sm := &opflexServiceMapping{
+			Conntrack:    true,
+		}
+		as.ServiceMappings = append(as.ServiceMappings, *sm)
+		agent.opflexServices[SnatService] = as
+		if !file_exists {
```
snaiksat: Not a big deal, but would you not be doing this check immediately after the L360 check? That way you don't do L362 to L377 processing if the file already exists.

snaiksat:
Also, as a follow up patch, it would be preferable to add a UT which checks for the file creation/deletion, and contents of the file when created.

pkg/hostagent/snats.go
snaiksat: @apoorva11029 for the offline explanation as to why its done this way. Basically the reason seems to be that we want the cache to be populated correctly and it wont be if the file is present (and the check is moved t

PR #314, opened 2019-08-29
Added single zone id for the cluster.

jayaramsatya:
this value can also be derived from configmap otherwise it will take default value


jayaramsatya:
> Changes look fine to me. Seems like editor introduced tab spaces?

Yeah previous checkin is not indented properly with my plugin it is corrected now

PR #318, opened 2019-09-06
launch-ovs.sh: fix postrotate script for ovs logrotation

akurz:
ovs container stops logging to STDOUT after the first logrotation
because the logfile is renamed and ovs is not informed to close/reopen
it's log. The reason is the postrotate script calling "ovs-appctl" with
empty targets:

ovs-appctl -t "" vlog/reopen

This fix prevents variable expansion and command substitution in the
HERE document that creates the logrotate configuration vor ovs.


snaiksat:
Thanks for catching this.

PR #320, opened 2019-09-09
IPSLA handling for SNAT

kishorejonnala:
When IPSLA is enabled (by configuring IPSLA interval to a >0 value), create one health-group per compute-node.
This will make sure that from fabric side, the packets are not hashed/black-holed towards the compute-nodes that are not reachable.

Fix has been tested on vk8s-2 where 3 of the nodes doesn't have any user-pods (ARP flows are not configured).
Also, tried shrink/expanding k8s cluster and observed that the health-group is cleaned-up accordingly.

### DIFF
--- a/pkg/controller/services.go
+++ b/pkg/controller/services.go
@@ -307,6 +307,37 @@ func apicRedirectPol(name string, tenantName string, nodes []string,
 	return rp, rpDn
 }
 
+func apicRedirectPolSnat(name string, tenantName string, nodes []string,
+	nodeMap map[string]*metadata.ServiceEndpoint,
+	monPolDn string, healthGroupDnMap map[string]string) (apicapi.ApicObject, string) {
+
+	rp := apicapi.NewVnsSvcRedirectPol(tenantName, name)
+	rp.SetAttr("thresholdDownAction", "deny")
+	rpDn := rp.GetDn()
+	for _, node := range nodes {
+		healthGroupDn, ok := healthGroupDnMap[node]
+		if !ok {
+			continue
+		}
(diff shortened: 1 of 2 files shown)


pkg/controller/services.go
```diff
@@ -660,20 +691,25 @@ func (cont *AciController) updateServiceDeviceInstanceSnat(key string) error {
                 // and IP address of each of the service endpoints for
                 // each node that hosts a pod for this service.  The
                 // example below shows the case of two nodes.
+				healthGroupDnMap := make(map[string]string)
                 var healthGroupDn string
+			    // For SNAT if IPSLA is enabled:
```
jojimt: Please run gofmt.

pkg/controller/services.go
```diff
@@ -307,6 +307,37 @@ func apicRedirectPol(name string, tenantName string, nodes []string,
...
 }
 
+func apicRedirectPolSnat(name string, tenantName string, nodes []string,
+	nodeMap map[string]*metadata.ServiceEndpoint,
+	monPolDn string, healthGroupDnMap map[string]string) (apicapi.ApicObject, string) {
+
+	rp := apicapi.NewVnsSvcRedirectPol(tenantName, name)
+	rp.SetAttr("thresholdDownAction", "deny")
+	rpDn := rp.GetDn()
+	for _, node := range nodes {
```
jojimt: I am wondering why we need to range over a nodes slice when we have two maps with nodes as keys. Can't we just range over one of those maps instead?

kishorejonnala: The reason to range over nodes slice is because healthGroupDn and ServiceEP are per node.

PR #325, opened 2019-09-13
Fix PBR-Tracking for services

kishorejonnala:
Current code created one health-group for all compute-nodes when IPSLA interval is set to a positive value.
This is incorrect, as health-group comes up when one of the compute-nodes are unreachable.

Fix is to create one HG per node and associate them to the redirect policy when IPSLA is configured to be more than zero.
With this fix, health-group's are created part of ServiceEndpoint creation (only when IPSLA is configured) and store healthgroupDn in ServiceEndPoint Metadata and use them when Redirect policy is created as the services comes-up.

Tested the behavior on vk8s testbed and there is enough UT coverage.

### DIFF
--- a/pkg/metadata/annotations.go
+++ b/pkg/metadata/annotations.go
@@ -32,9 +32,10 @@ type OpflexGroup struct {
 
 // annotation type for service endpoint information
 type ServiceEndpoint struct {
-	Mac  string `json:"mac,omitempty"`
-	Ipv4 net.IP `json:"ipv4,omitempty"`
-	Ipv6 net.IP `json:"ipv6,omitempty"`
+	HealthGroupDn   string `json:"-"`
+	Mac  			string `json:"mac,omitempty"`
+	Ipv4 			net.IP `json:"ipv4,omitempty"`
+	Ipv6 			net.IP `json:"ipv6,omitempty"`
 }
 
 // annotation type for IPs allocation chunks
--- a/pkg/controller/cf_environment.go
+++ b/pkg/controller/cf_environment.go
@@ -672,7 +672,7 @@ func (env *CfEnvironment) SetCellServiceInfo(nodeName, cellId string) {
 		existing = &currMeta.serviceEp
 	}
 	err := env.cont.createServiceEndpoint(existing, &nodeMeta.serviceEp,
-		deviceMac)
+		deviceMac, nodeName)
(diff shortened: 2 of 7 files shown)


pkg/controller/controller.go
```diff
@@ -328,6 +328,13 @@ func (cont *AciController) Run(stopCh <-chan struct{}) {
 		panic(err)
 	}
 
+	// If not defined, default to 6
+	if cont.config.AciServiceMonitorInterval == 0 {
+		cont.config.AciServiceMonitorInterval = 6
+	}
+	cont.log.Info("AciServiceMonitorIntervalPodIpPoolChunkSize conf is set to: ", cont.config.AciServiceMonitorInterval)
+
+
```
kishorejonnala: This is added mainly for testing; I will remove before merge.

jojimt: Please set to Debug, instead of removing it.

pkg/metadata/annotations.go
```diff
@@ -32,9 +32,10 @@ type OpflexGroup struct {
 
 // annotation type for service endpoint information
 type ServiceEndpoint struct {
-	Mac  string `json:"mac,omitempty"`
-	Ipv4 net.IP `json:"ipv4,omitempty"`
-	Ipv6 net.IP `json:"ipv6,omitempty"`
+	HealthGroupDn   string `json:"health-group,omitempty"`
```
jojimt: I think we should use json:"-" for HealthGroupDn as we don't need to write this as json.

pkg/controller/controller.go
kishorejonnala: I meant to say - I will remove this code of setting IPSLA value to positive number with-in ACC since this is exposed from acc-provision and we need to honor the setting coming from user.

pkg/metadata/annotations.go
kishorejonnala: apicRedirectDst() relys on this to be an empty string for cases where IPSLA is not enabled. Do you see an issue by leaving it this way?

jojimt: That shouldn't be affected. This matters only when we call json.Marshal directly on this struct.

kishorejonnala: ok, I just took care of it.

PR #327, opened 2019-09-13
Fix acikubectl exec command to pick acc container

apoorva11029:


cmd/acikubectl/cmd/version.go
```diff
@@ -25,33 +25,43 @@ import (
 )
 
 func getVersion(cmd *cobra.Command, args []string) {
+	fmt.Println("CHECK 0")
```
snaiksat: remove the debug prints?

PR #331, opened 2019-09-16
Fix tagAnnotation dn format

kishorejonnala:
It appears that APIC4.1 onwards seems to have changed the format of tagAnnotation object.
This change will accommodate the same, so that it avoids treating an object as altered when processing a websocket notification on any of the ACC managed objects.

Example:
Prior to 4.1: uni/tn-common/brc-vk8s_2_snat_svcgraph/annotationKey-aci-containers-controller-tag
4.1+        : uni/tn-common/brc-vk8s_2_snat_svcgraph/annotationKey-[aci-containers-controller-tag]

Tests:
Reproduced the issue on Fab14 and tested with fix.

### DIFF
--- a/pkg/apicapi/apic_types.go
+++ b/pkg/apicapi/apic_types.go
@@ -487,10 +487,15 @@ func NewTagInst(parentDn string, name string) ApicObject {
 }
 
 func NewTagAnnotation(parentDn string, key string) ApicObject {
+	dn := ""
 	ret := newApicObject("tagAnnotation")
 	ret["tagAnnotation"].Attributes["key"] = key
-	ret["tagAnnotation"].Attributes["dn"] =
-		fmt.Sprintf("%s/annotationKey-%s", parentDn, key)
+	if ApicVersion >=4.1 {
+		dn = fmt.Sprintf("%s/annotationKey-[%s]", parentDn, key)
+	} else {
+		dn = fmt.Sprintf("%s/annotationKey-%s", parentDn, key)
+	}
+	ret["tagAnnotation"].Attributes["dn"] = dn
 	return ret
 }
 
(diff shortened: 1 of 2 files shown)


pkg/apicapi/apic_types.go
```diff
@@ -490,7 +490,7 @@ func NewTagAnnotation(parentDn string, key string) ApicObject {
 	ret := newApicObject("tagAnnotation")
 	ret["tagAnnotation"].Attributes["key"] = key
 	ret["tagAnnotation"].Attributes["dn"] =
-		fmt.Sprintf("%s/annotationKey-%s", parentDn, key)
+		fmt.Sprintf("%s/annotationKey-[%s]", parentDn, key)
```
snaiksat: Is this change backward compatible (with say 3.2)?

kishorejonnala: While there is no functional impact, I went ahead and made this version dependent to handle it cleanly.

pkg/apicapi/apicapi.go
```diff
@@ -44,6 +44,13 @@ import (
 // RefreshInterval is set to 0
 const defaultConnectionRefresh = 30 * time.Second
 
+// ApicVersion - This global variable to be used when dealing with version-
+// dependencies during APIC interaction. It gets filled with actual version
+// as part of runConn()
+var (
+	ApicVersion = 3.1
```
snaiksat: If we are initializing here, it should probably be 3.2 since we don't support 3.1.

pkg/apicapi/apic_types.go
snaiksat: Thanks. There was no functional impact to begin with. We still decided to make the change in favor of the newer release and but introduced the same issue in the older release. Given that we are just as likely to have the older release (3.2) deployed at least for some time, preserving backward compatibility makes sense.

PR #337, opened 2019-09-18
Null mac ep file is created default 

jayaramsatya:
Null Mac EP file will be created when hostagent comes up. previously we used to create based snat policy info.


snaiksat:
It would have been good to note in the commit message that we don't delete this file any more.

PR #381, opened 2020-01-08
Allow SNAT with services exposing multiple ports

apoorva11029:
Fix for issue https://cdetsng.cisco.com/webui/#view=CSCvs41252


snaiksat:
Can we have a test in acc-pytest for this?

PR #401, opened 2020-01-27
Collect configmaps in cluster-report

apoorva11029:
1. Save aci-containers-config cmap in cluster-report/status/configmap_controller.log
2. Save snat operator cmaps(snat-operator-config and snat-operator-lock) in cluster-report/status/configmap_snatoperator.log

Also removed the flag --include-uninitialized from kubectl get commands as it's no longer supported starting k8s1.16. Now kubectl will show all uninitialized resources by default itself.

(cherry picked from commit 87fce5712b04dab41a259a322ba5a1e25eea7a08)


snaiksat:
We will create a new branch so no need to backport.

PR #403, opened 2020-01-28
Filter services for snat by namespace: bug fix

apoorva11029:
Fix for CSCvs41278

### DIFF
--- a/pkg/controller/environment.go
+++ b/pkg/controller/environment.go
@@ -26,7 +26,6 @@ import (
 	"github.com/yl2chen/cidranger"
 	v1 "k8s.io/api/core/v1"
 	v1net "k8s.io/api/networking/v1"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/kubernetes"
 	restclient "k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/cache"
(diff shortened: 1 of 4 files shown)


pkg/controller/environment.go
```diff
@@ -132,8 +132,8 @@ func (env *K8sEnvironment) Init(cont *AciController) error {
 			Services(service.ObjectMeta.Namespace).UpdateStatus(service)
 	}
 
-	cont.listServicesBySelector = func(selector string) (*v1.ServiceList, error) {
-		return kubeClient.CoreV1().Services("").List(metav1.ListOptions{LabelSelector: selector})
+	cont.listServicesBySelector = func(labelSelector string, fieldSelector string) (*v1.ServiceList, error) {
+		return kubeClient.CoreV1().Services("").List(metav1.ListOptions{LabelSelector: labelSelector, FieldSelector: fieldSelector})
```
jayaramsatya: Can you list the services from local store right? do you need to make a client call here ? otherwise  changes looks good

apoorva11029: Done!

pkg/controller/snatglobalinfo.go
```diff
@@ -316,20 +315,11 @@ func (cont *AciController) deleteNodeinfoFromGlInfoCache(nodename string, global
...
-		func(servobj interface{}) {
-			services = append(services, servobj.(*v1.Service))
-		})
+        services := cont.getServicesBySelector(labels.SelectorFromSet(
+                            labels.Set(policy.Selector.Labels)),
+                                    policy.Selector.Namespace)
 	for _, service := range services {
-		if len(policy.Selector.Namespace) == 0 || (len(policy.Selector.Namespace) > 0 &&
-			policy.Selector.Namespace == service.ObjectMeta.Namespace) {
-			if len(service.Status.LoadBalancer.Ingress) == 0 {
```
jayaramsatya: This check is required  please put the changes back

jayaramsatya: if len(service.Status.LoadBalancer.Ingress) == 0 {

apoorva11029: As discussed offline, this check is in the new function to list services

PR #406, opened 2020-01-30
ACI CNI Operator release for OSP-4.3/K8s platform

abhis2112:
Initial Commit for ACI CNI Operator for OSP-4.3/K8s platform


cmd/acioperator/main.go
```diff
@@ -0,0 +1,113 @@
...
+	log.Debug("Initializing kubernetes client")
+	client := getOperatorClient()
+
+	informer := cache.NewSharedIndexInformer(
+		&cache.ListWatch{
+			ListFunc: func(options meta_v1.ListOptions) (runtime.Object, error) {
+				return client.AciV1alpha1().AciContainersOperators("aci-containers-system").List(options)
+			},
+			WatchFunc: func(options meta_v1.ListOptions) (watch.Interface, error) {
+				return client.AciV1alpha1().AciContainersOperators("aci-containers-system").Watch(options)
```
jayaramsatya: Can you change this hardcoded value "aci-containers-system" and also check if you can read it from environment value from deployment file

cmd/acioperator/main.go
```diff
@@ -0,0 +1,113 @@
...
+			},
+		},
+		&operators.AciContainersOperator{},
+		0,
+		cache.Indexers{},
+	)
+
+	queue := workqueue.NewRateLimitingQueue(workqueue.DefaultControllerRateLimiter())
+
+	informer.AddEventHandler(cache.ResourceEventHandlerFuncs{
```
jayaramsatya: this cmd file right you can register the event handlers in  operator file right, the way we are following for controller and hostagent. if you want to handle it different commit that is fine

pkg/acioperator/operator.go
```diff
@@ -0,0 +1,371 @@
...
+		err = rclient.Update(ctx, clusterConfig)
+		if err != nil {
+			log.Info(err)
+			return
+		}
+	}
+
+
+
+	//err := client.Get(context.TODO(), types.NamespacedName{
```
jayaramsatya: can you . remove this code if you are not using this code

abhis2112:
Redo

pkg/acioperator/operator.go
```diff
@@ -0,0 +1,371 @@
...
+func (c *Controller) processNextItem() bool {
+	log.Info("Controller.processNextItem: starting....")
+
+	key, quit := c.Queue.Get()
+
+	if quit {
+		return false
+	}
+
+	defer c.Queue.Done(key)
```
jayaramsatya: do you need to shutdown the queue here ?

pkg/acioperator/operator.go
```diff
@@ -0,0 +1,371 @@
...
+		} else {
+			c.Logger.Errorf("Controller.processNextItem: Failed to process the item with key %s with error %v, no more retries", key, err)
+			c.Queue.Forget(key)
+			utilruntime.HandleError(err)
+		}
+	}
+
+	// if the item doen not exist means deleted otherwise it is created or updated.
+	// After processing we will remove the key from queue
+	if !exists {
```
jayaramsatya: check for item also whether item is nil add the or condition here

pkg/acioperator/operator.go
```diff
@@ -0,0 +1,371 @@
...
+		return
+	}
+
+	cmd := exec.Command("kubectl","--token=" + string(token),"apply","-f","manifests")
+	log.Println(cmd)
+	stdout, err := cmd.Output()
+
+	log.Info(string(stdout))
+
+	k8sclient := getk8sClient()
```
jayaramsatya: Need to check for the k8sclient is nil

pkg/acioperator/operator.go
```diff
@@ -0,0 +1,371 @@
...
+	log.Info("Applying Deployment")
+	token, err := ioutil.ReadFile("/var/run/secrets/kubernetes.io/serviceaccount/token")
+	if err != nil {
+		fmt.Println(err)
+		return
+	}
+
+	cmd := exec.Command("kubectl","--token=" + string(token),"apply","-f","manifests")
+	log.Println(cmd)
+	stdout, err := cmd.Output()
```
jayaramsatya: check for error condition here

pkg/acioperator/operator.go
```diff
@@ -0,0 +1,371 @@
...
+					*metav1.NewControllerRef(acicontainersoperator, operators.SchemeGroupVersion.WithKind("AciContainersOperator")),
+				}
+
+	_, err = deploymentsClient.Update(t.Deployment)
+	if err != nil {
+		fmt.Println(err.Error())
+	}
+
+
+	hostdaemonsetclient := k8sclient.AppsV1().DaemonSets("aci-containers-system")
```
jayaramsatya: you need to validate the client before accessing it

pkg/acioperator/operator.go
```diff
@@ -0,0 +1,371 @@
...
+		}
+		log.Info("Config Spec is  ", clusterConfig.Spec)
+		clusterConfig.Status.ClusterNetwork = clusterConfig.Spec.ClusterNetwork
+		clusterConfig.Status.NetworkType = clusterConfig.Spec.NetworkType
+		clusterConfig.Status.ServiceNetwork = clusterConfig.Spec.ServiceNetwork
+
+		log.Info("Status is ", clusterConfig.Status)
+
+		ctx := context.TODO()
+		err = rclient.Update(ctx, clusterConfig)
```
jayaramsatya: if your update fails you need to requeue the request here. there is a chance that etcd writes can fail here

PR #412, opened 2020-02-01
Openshift system namespaces annotation restriction

apoorva11029:
Don't allow the following namespaces to be annotated with "opflex.cisco.com/endpoint-group" with name "kubernetes|kube-system" for any new OpenShift clusters: kube-service-catalog,kube-service-catalog, openshift-console, openshift-monitoring and openshift-web-console

This change is made to handle the new acc-provision resources naming convention starting Jordan. 

Also blocks our openshift-ansible installation scripts <= 3.11 from setting the eg group to "kubenetes|kube-system" (https://github.com/noironetworks/openshift-ansible/commit/0ad747948460a3fdd7b57d49c0a43a20eda9686d)

### DIFF
--- a/pkg/hostagent/group_assign.go
+++ b/pkg/hostagent/group_assign.go
@@ -186,9 +186,26 @@ func (agent *HostAgent) assignGroups(pod *v1.Pod) (metadata.OpflexGroup, []metad
 	var namespace *v1.Namespace
 	if exists && namespaceobj != nil {
 		namespace = namespaceobj.(*v1.Namespace)
+                var temp metadata.OpflexGroup
+                decodeAnnotation(namespace.ObjectMeta.Annotations[metadata.EgAnnotation], &temp, logger, "namespace[EpgAnnotation]")
 
-		decodeAnnotation(namespace.ObjectMeta.Annotations[metadata.EgAnnotation], &egval, logger, "namespace[EpgAnnotation]")
-		decodeAnnotation(namespace.ObjectMeta.Annotations[metadata.SgAnnotation], &sgval, logger, "namespace[SgAnnotation]")
+                // For OpenShift installations <= 3.11, we want to offset
+                // the namespace annotation applied in this PR:
+                // openshift/openshift-ansible@0ad7479. The pods
+                // belonging to the four namespaces (i.e.
+                // kube-service-catalog,kube-service-catalog,
+                // openshift-console, openshift-monitoring and
+                // openshift-web-console) need to go into the system
+                // EPG, which we do in acc-provision:
+                // noironetworks/acc-provision@f93a132
+                if (agent.config.AciVmmDomainType == "OpenShift" && temp.Name == "kubernetes|kube-system" &&
+                           (namespace.ObjectMeta.Name == "kube-service-catalog" || namespace.ObjectMeta.Name == "openshift-console" ||
+                                    namespace.ObjectMeta.Name == "openshift-monitoring" || namespace.ObjectMeta.Name == "openshift-web-console")){
+                                        agent.log.Debug("Ignoring annotation set on ", namespace.ObjectMeta.Name)
+                } else {
+                        egval = temp
+                }
+                decodeAnnotation(namespace.ObjectMeta.Annotations[metadata.SgAnnotation], &sgval, logger, "namespace[SgAnnotation]")
 	}
 
 	// annotation on parent deployment or rc is next-highest priority


snaiksat:
1. Just to confirm, if OpenShift cluster (<= 3.11) is reinstalled, the changes in https://github.com/noironetworks/acc-provision/pull/167/files#diff-b4056411436ca3895fed0f4dc6300d09R520-R524 will take care of identifying the appropriate OpenShift namespaces from which to put the pods in the aci-containers-system EPG?
2. Also, for existing clusters, as long as the annotations on any of the identified namespaces stays, their pods will remain in the kube-system EPG and things will continue to keep working?

jojimt:
Related question - what will the user see in terms of an error?

snaiksat:
@jojimt If your question is why we are doing this - it's because we can no longer modify upstream openshift-ansible changes done in this PR:
https://github.com/openshift/openshift-ansible/commit/0ad747948460a3fdd7b57d49c0a43a20eda9686d
Hence, for any new OpenShift 3.11 deployments we are offsetting the effect of the changes in that commit with this change, and also the corresponding changes in acc-provision (referred to earlier).

jojimt:
Checked the code. I see that we are ignoring the annotation. I think that would be very confusing from a user interface point of view. We should somehow indicate an error, perhaps by moving the pod into a quarantine epg.

snaiksat:
@jojimt - the annotation we are offsetting here is not a user annotation; it's an annotation we add through our ansible code during OpenShift install 3.11 (note, its only that specific annotation we ignore).  The pods belonging to the identified namespaces (i.e. kube-service-catalog,kube-service-catalog, openshift-console, openshift-monitoring and openshift-web-console) need to go into a specific EPG, which we correctly do by virtue of the changes already done in acc-provision. It would not make sense for the user to apply the annotation with "kubernetes|kube-system" because the app-profile being referred to, "kubernetes" would not even be present in the new naming convention (that profile is managed by us via acc-provision).

I see that we are logging that the particular annotation is being ignored so we are good.

The objective of this patch is not to generically prevent the user from applying annotations on just these namespaces. That's a separate discussion/patch if we want to do that, and would need to be done for other namespaces (not just these particular openshift namespaces).

jojimt:
I understand what you're saying. But this is a pretty nasty hack. Have we looked at other options?

snaiksat:
Hack, yes, but nasty?? It's beyond our control to change the upstream (that branch is closed). So there aren't too many other options. We can discuss further, but we need to need a fix at least until then.

jojimt:
The issue is it introduces obscurity into a primary user interface(not just the code). Just try describing the group assignment behavior before and after this change to see my point. Instead, can we simply provide a command to remove the annotation from the namespace using kubectl or oc? That seems like a more reasonable approach to me.

snaiksat:
That group annotation feature does not change, nor does the user interface to it. The above is an internal change, so there is no need to document this from a group annotation perspective. What we will document is the new naming convention, which applies to clusters deployed post the release of the new naming convention. The user does not see any change (either old cluster or new cluster).

Requiring to remove the annotation manually breaks the install. Not to mention that, it is actually something we will have to document (and make sure its propagated to every place where it was documented including upstream). Not very reasonable, compared to the mostly transparent change happening in this patch.

jojimt:
I think there is a disconnect in understanding here. The code clearly shows the group assignment behavior is changed. You're saying it has not. Let's discuss this in person.

pkg/hostagent/group_assign.go
```diff
@@ -186,9 +186,16 @@ func (agent *HostAgent) assignGroups(pod *v1.Pod) (metadata.OpflexGroup, []metad
 	var namespace *v1.Namespace
 	if exists && namespaceobj != nil {
 		namespace = namespaceobj.(*v1.Namespace)
-
-		decodeAnnotation(namespace.ObjectMeta.Annotations[metadata.EgAnnotation], &egval, logger, "namespace[EpgAnnotation]")
-		decodeAnnotation(namespace.ObjectMeta.Annotations[metadata.SgAnnotation], &sgval, logger, "namespace[SgAnnotation]")
+                var temp metadata.OpflexGroup
+                decodeAnnotation(namespace.ObjectMeta.Annotations[metadata.EgAnnotation], &temp, logger, "namespace[EpgAnnotation]")
```
jojimt: Please make the vmmType check the first level check so it's clear it's applicable to only openshift. Also, please add a comment.

apoorva11029: Added, thanks!

PR #413, opened 2020-02-03
veth_host fixes

mchalla:
- patch ports should be added to ignore list
  always (first time and after host agent restart)
- avoid creation of duplicate iptables / nft rules
- tested by restarting host-agent and
  rebooting the node

Signed-off-by: Madhu Challa <challa@gmail.com>

### DIFF
--- a/docker/launch-hostagent.sh
+++ b/docker/launch-hostagent.sh
@@ -73,24 +73,33 @@ fi
 if [[ ! -z "$VTEP_IFACE" && ! -z "$VTEP_IP" ]]; then
 
     set +e
-    # Allow pod traffic to go out of veth_host
-    iptables -A FORWARD -i veth_host -j ACCEPT
-    retval=$?
-    if [ $retval -ne 0 ]; then
-        echo "command failed, trying with nftables"
-        nft add table ip filter
-        nft add chain ip filter FORWARD
-        nft add rule ip filter FORWARD iifname veth_host counter accept
-    fi
 
-    # SNAT outgoing traffic from pod to external world
-    iptables -t nat -A POSTROUTING -o $VTEP_IFACE -j MASQUERADE
+    iptables-save
     retval=$?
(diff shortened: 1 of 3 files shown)


docker/launch-hostagent.sh
```diff
@@ -73,24 +73,33 @@ fi
...
+            nft add chain ip filter FORWARD
+            nft add rule ip filter FORWARD iifname veth_host counter accept
+
+            # SNAT outgoing traffic from pod to external world
+            nft add table ip nat
+            nft add chain ip nat POSTROUTING
+            nft add rule ip nat POSTROUTING oif $VTEP_IFACE masquerade
+        fi
+    else
+        iptables-save | grep veth_host
```
snaiksat: nit: as an enhancement we can perhaps make this check more specific.

docker/launch-opflexserver.sh
```diff
@@ -26,4 +26,4 @@ if [ ! -f ${OPFLEXSERVER_CONF} ]; then
 EOF
 fi
 
-exec ${OPFLEXSERVER} --policy=${OPFLEXSERVER_POL} --grpc_conf=${OPFLEXSERVER_CONF} $@
+exec ${OPFLEXSERVER} --level=debug2 --policy=${OPFLEXSERVER_POL} --grpc_conf=${OPFLEXSERVER_CONF} $@
```
snaiksat: is this intentional?

docker/launch-hostagent.sh
mchalla: i was going to, but it was getting complicated. Since we add both rules at same place they should either both be present or not and i felt that one check is simple. I also checked the -C flag and some versions of iptables do not have the flag.

docker/launch-opflexserver.sh
mchalla: yes. without this no logs are printed for opflex server and its not clear if its hung or processing anything.

PR #415, opened 2020-02-05
Minor fix for pod group assign

apoorva11029:


pkg/hostagent/group_assign.go
```diff
@@ -203,7 +203,7 @@ func (agent *HostAgent) assignGroups(pod *v1.Pod) (metadata.OpflexGroup, []metad
                                     namespace.ObjectMeta.Name == "openshift-monitoring" || namespace.ObjectMeta.Name == "openshift-web-console")){
                                         agent.log.Debug("Ignoring annotation set on ", namespace.ObjectMeta.Name)
                 } else {
-                        egval = temp
+                        decodeAnnotation(namespace.ObjectMeta.Annotations[metadata.EgAnnotation], &egval, logger, "namespace[EpgAnnotation]")
```
jayaramsatya: this will override the default value of eval if the annotations length is zero. make sure that you have valid annotations

apoorva11029: If the annotation length is zero, we don't modify the egval. Ref: https://github.com/noironetworks/aci-containers/blob/aafeb5a31d7b708d5cf6ce2b409f333cceb78a21/pkg/hostagent/group_assign.go#L100
Is that what you meant?

jayaramsatya: yeah, correct I didn't check it. it should take care of it

snaiksat: The check seems to be happening in that function.

PR #421, opened 2020-02-08
[ACI-Istio-Operator]: Changes to drive istio control plane install through ACI-CNI

kishorejonnala:
Summary of the changes:
- Operator framework based ACIIstioOperator to drive Istio control plane install through ACI-CNI
- AciIstioOperatorSpec is defined with two desired state params (Config/Profile) with the assumption that "config" will carry upstream istio-operator deployment file and "profile" will determine which istio control plane profile to be installed on the cluster.
- ACC container to create the AciIstioCR and trigger upstream istio-operator deployment (upstream-istio-operator.yaml and upstream-istio-ctrlplane-resource.yaml)
- pkg/controller/aci_istio.go has the busniess logic to handle Create/Delete of aciistiooperator. 
- acc-provision changes [Commit](https://github.com/noironetworks/acc-provision/commit/25186ca21ce9e1b8f206bf608e8e1ef176467937) will annotate istio-operator and istio-system Namespaces with kube-istio EPG to provide the isolation.

Workflow:
- ACC-Provision  include aciistiooperator (CRD) in the deployment file based on the config knob "istio_config":"install_istio" (default set to True ). 

- When deployed, AciIstioOperator backend logic will create a CR and deploy both upstream-istio-operator.yaml and upstream-istio-ctrlplane-resource.yaml. Also, mark the parent reference to "istio-operator" deployment.

- When AciIstioOperator is deleted, "istio-operator" deployment will be deleted as well due to the parent referencing. It also cleans up "istio-system" namespace upon AciIstioOperator deletion.

ToDo:
- Handling the update to "profile" parameter to the AciIstioOperator spec.


pkg/controller/aci_istio.go
```diff
@@ -0,0 +1,294 @@
...
+	"k8s.io/client-go/util/workqueue"
+	"k8s.io/apimachinery/pkg/watch"
+	"os"
+	"os/exec"
+	"time"
+)
+
+// Controller here defines the Operator code handler which list watch the AciIstioOperator
+// object and apply istio-deployment.yaml in the cluster after creation/updation
+type AciIstioController struct {
```
apoorva11029: nit: ACI should be spelled in caps as per golang convention. We should also do the switch in our other packages, but maybe we could start with istio?

kishorejonnala: Sure, I will rename with my next commit.

docker/Dockerfile-aci-istio-operator
```diff
@@ -0,0 +1,16 @@
+FROM alpine:3.10.2
+ENV http_proxy=http://proxy.esl.cisco.com:80
+ENV https_proxy=http://proxy.esl.cisco.com:80
```
snaiksat: I guess we need to remove this. ;-)

docker/Dockerfile-aci-istio-operator
```diff
@@ -0,0 +1,16 @@
...
+RUN apk upgrade --no-cache
+RUN apk update && apk add curl git
+RUN curl -LO https://storage.googleapis.com/kubernetes-release/release/`curl -s https://storage.googleapis.com/kubernetes-release/release/stable.txt`/bin/linux/amd64/kubectl
+#RUN curl -LO https://storage.googleapis.com/kubernetes-release/release/v1.14.6/bin/linux/amd64/kubectl
```
snaiksat: please remove if not needed...

cmd/aciistio/main.go
```diff
@@ -0,0 +1,64 @@
+// Copyright 2016 Cisco Systems, Inc.
```
snaiksat: nit: 2020

pkg/controller/aci_istio.go
```diff
@@ -0,0 +1,294 @@
...
+	}
+
+	defer c.Queue.Done(key)
+
+	keyRaw := key.(string)
+
+	item, exists, err := c.Informer.GetIndexer().GetByKey(keyRaw)
+	if err != nil {
+		if c.Queue.NumRequeues(key) < 10 {
+			c.Logger.Errorf("Controller.processNextItem: Failed to process the item with key %s with error %v, retrying", key, err)
```
snaiksat: nit: I am not sure we want to mention "Controller.processNextItem" since i believe the logger will add, however if we do want to perhaps change the name to AciIstioController

docker/Dockerfile-aci-istio-operator
```diff
@@ -0,0 +1,16 @@
...
+RUN chmod u+x kubectl && mv kubectl /usr/local/bin/kubectl
+RUN curl -LO https://istio.io/operator.yaml
+RUN mkdir -p /usr/local/manifests
```
snaiksat: nit: For consistency, this could be in the /usr/local/var/lib/ path (we have other directories there like opflex-agent).

pkg/controller/aci_istio.go
```diff
@@ -289,12 +273,22 @@ func (t *AciIstioOperatorHandler) ObjectCreated(obj interface{}) {
...
 }
 
 func (t *AciIstioOperatorHandler) ObjectDeleted(obj interface{}) {
-	log.Info("OperatorHandler.ObjectDeleted")
+
+	log.Info("Deleting istio-system namespace")
+	cmd := exec.Command("kubectl", "delete", "ns", "istio-system",  "--grace-period=0", "--force")
+	out, err := cmd.Output()
+	if err != nil {
+		log.Info("Failed:",err)
```
jayaramsatya: Can you return the code form here otherwise both success and Failed message will be seen

pkg/controller/aci_istio.go
```diff
@@ -289,12 +273,22 @@ func (t *AciIstioOperatorHandler) ObjectCreated(obj interface{}) {
...
+	}
+	log.Info("Success:", string(out))
```
jayaramsatya: Also check if we can move to Debug Level

docker/Dockerfile-aci-istio-operator
kishorejonnala: @snaiksat I kept it so that my local build wouldn't fail every time. I understand the Jekins would take care of the proxy settings, would it hurt to keep them here? They are reset at end of file. If I remove them, I will have to add these every time I need to build locally.

docker/Dockerfile-aci-istio-operator
kishorejonnala: Done. New path: /usr/local/var/lib/aci-cni/

pkg/controller/aci_istio.go
kishorejonnala: Fixed.

pkg/controller/aci_istio.go
kishorejonnala: Done.

docker/Dockerfile-aci-istio-operator
```diff
@@ -0,0 +1,15 @@
...
+ENV https_proxy=http://proxy.esl.cisco.com:80
+RUN apk upgrade --no-cache
+RUN apk update && apk add curl git
+RUN curl -LO https://storage.googleapis.com/kubernetes-release/release/`curl -s https://storage.googleapis.com/kubernetes-release/release/stable.txt`/bin/linux/amd64/kubectl
+RUN chmod u+x kubectl && mv kubectl /usr/local/bin/kubectl
+RUN curl -LO https://istio.io/operator.yaml
+RUN mkdir -p /usr/local/var/lib/aci-cni
+RUN mv operator.yaml /usr/local/var/lib/aci-cni/upstream-istio-operator.yaml 
+RUN echo -e "apiVersion: install.istio.io/v1alpha2\nkind: IstioControlPlane\nmetadata:\n  namespace: istio-operator\n  name: example-istiocontrolplane\nspec:\n  profile: demo\n---" >> /usr/local/var/lib/aci-cni/upstream-istio-ctrlplane-resource.yaml
+COPY dist-static/aci-istio-operator /usr/local/bin/
```
snaiksat: follow up PR suggestion - Can we use demo as default but also have an option in config map to override this?

pkg/controller/aci_istio.go
```diff
@@ -0,0 +1,345 @@
...
+
+func (c *AciIstioController) CreateAciIstioOperatorCR() error{
+
+	aciIstioCR := &istiooperators.AciIstioOperator{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: "aciistiooperator",
+			Namespace: os.Getenv("SYSTEM_NAMESPACE"),
+		},
+		Spec: istiooperators.AciIstioOperatorSpec{
+			Profile: "demo",
```
snaiksat: Per earlier comment, this can go in configmap.

PR #426, opened 2020-02-14
CSCvq96281 Validating Vrf config

kishorejonnala:
- This fix introduces a validation to make sure that POD/NodeBD/L3Out are pointing to the same VRF that's specified by the user in acc-provision input file.
- Fix is verified on Fab10

### DIFF
--- a/pkg/apicapi/apicapi.go
+++ b/pkg/apicapi/apicapi.go
@@ -29,6 +29,7 @@ import (
 	"net/http/cookiejar"
 	"regexp"
 	"strings"
+	"sort"
 	"time"
 	"strconv"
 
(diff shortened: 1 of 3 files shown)


snaiksat:
Preferably add a UT in a follow up patch to catch the error case.

pkg/apicapi/apicapi.go
```diff
@@ -746,6 +747,61 @@ func (conn *ApicConnection) logErrorResp(message string, resp *http.Response) {
...
+		return err
+	}
+
+    for _, obj := range apicresp.Imdata {
+        for _, body := range obj {
+			tDn,ok := body.Attributes["tDn"].(string)
+			if !ok {
+				continue
+			}
+			aciVrfBdL3OuttDns = append(aciVrfBdL3OuttDns, tDn)
```
jayaramsatya: can you check if you can move the below for loop to here

kishorejonnala:
@snaiksat  Sure, will add a UT as a follow-up patch.

pkg/apicapi/apicapi.go
kishorejonnala: @jayaramsatya I don't like too many loops as well. In this case, APIC can always have a superset of tDns and hence it's not possible to compare and error out. We can do it by having a map of expectedDns and marking them for each tDn that we read from APIC, but it gets unreadable. Given that we are only dealing with handful (3 as of now) of expectedDns, it's ok to have that additional loop in the end.

PR #428, opened 2020-02-15
Final fix for the K8s/OpenShift for Config-Map/kubeconfig/OSP and the application of Manifests

abhis2112:
The PR includes following fix:
1. Config Map handler to create CR
2. Removal of Kubeconfig File dependency
3. Change of manifest file name inside the operator pod to aci-deployment.yaml
4. Openshift changes to update network status


pkg/controller/acicontainersoperator.go
```diff
@@ -353,48 +408,57 @@ func (t *OperatorHandler) ObjectCreated(obj interface{}) {
...
+		}
+
+		err = rclient.Get(context.TODO(), types.NamespacedName{
+			Name: "cluster",
+		}, clusterConfig)
+		if err != nil {
+			log.Info(err)
+		}
+
+		log.Info("Config Spec is  ", clusterConfig.Spec)
```
snaiksat: If this is too verbose, pl change it to debug

pkg/controller/acicontainersoperator.go
```diff
@@ -130,6 +130,49 @@ func NewAciContainersOperator(
...
+	var options metav1.GetOptions
+	_,er := c.Clientset.AciV1alpha1().AciContainersOperators(os.Getenv("SYSTEM_NAMESPACE")).Get("acicnioperator",options)
+	if er != nil{
+		return er
+	}
+	return nil
+}
+
+func (c *Controller) CreateAciContainersOperatorCR() error{
+	log.Info("Reading the Config Map providing CR")
```
snaiksat: Should this be info level?

pkg/controller/acicontainersoperator.go
```diff
@@ -130,6 +130,49 @@ func NewAciContainersOperator(
...
+	obj := &operators.AciContainersOperator{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: "acicnioperator",
+			Namespace: os.Getenv("SYSTEM_NAMESPACE")},
+	}
+
+	log.Info("Unmarshalling the Config-Map...")
+	err = json.Unmarshal(raw, &obj.Spec)
+	if err != nil{
+		return err
```
snaiksat: It might be helpful to log this at info.

pkg/controller/acicontainersoperator.go
```diff
@@ -130,6 +130,49 @@ func NewAciContainersOperator(
...
+	}
+	log.Debug("CR object recieved is",(obj.Spec))
```
snaiksat: It might be helpful to log at info that the unmarshalling was successful.

pkg/controller/acicontainersoperator.go
```diff
@@ -138,6 +181,22 @@ func (c *Controller) Run(stopCh <-chan struct{}) {
 
 	c.Logger.Info("Controller.Run: initiating")
 
+	log.Info("Checking if Custom Resource already present")
+	err := c.GetAciContainersOperatorCR()
+	if err != nil {
+		log.Info("Not Present ..Creating Custom Resource")
```
snaiksat: Can you add the name of the custom resource (we know there is only one, but it will still make it more readable in the logs).

pkg/controller/acicontainersoperator.go
```diff
@@ -138,6 +181,22 @@ func (c *Controller) Run(stopCh <-chan struct{}) {
...
+		err = c.CreateAciContainersOperatorCR()
+		if err != nil {
+			log.Error("Function returned error",err)
```
snaiksat: By itself this log is not helpful, suggest saying that the CR creation failed.

pkg/controller/acicontainersoperator.go
```diff
@@ -138,6 +181,22 @@ func (c *Controller) Run(stopCh <-chan struct{}) {
...
+			return
+		}
+	}
+
+        if err == nil {
+
+                log.Info("Custom Resource already present")
```
snaiksat: Is it possible that c.GetAciContainersOperatorCR() fails for some other reason?

pkg/controller/acicontainersoperator.go
```diff
@@ -241,7 +299,9 @@ func (t *OperatorHandler) ObjectCreated(obj interface{}) {
 	log.Info("OperatorHandler.ObjectCreated")
```
snaiksat: Can this log be made more descriptive?

pkg/controller/acicontainersoperator.go
```diff
@@ -241,7 +299,9 @@ func (t *OperatorHandler) ObjectCreated(obj interface{}) {
...
 
 	acicontainersoperator := obj.(*operators.AciContainersOperator)
-	log.Info(acicontainersoperator.Spec.Config)
+
+	log.Debug(acicontainersoperator.Spec.Config)
```
snaiksat: Can you add a message before logging the config?

pkg/controller/acicontainersoperator.go
```diff
@@ -270,15 +330,10 @@ func (t *OperatorHandler) ObjectCreated(obj interface{}) {
 		return
 	}
 	log.Info("Applying Deployment")
```
snaiksat: You can consider lgging the deployment file at debug log level.

pkg/controller/acicontainersoperator.go
```diff
@@ -353,48 +408,57 @@ func (t *OperatorHandler) ObjectCreated(obj interface{}) {
...
-	//	log.Info("Status is ", clusterConfig.Status)
-	//
-	//	ctx := context.TODO()
-	//	err = rclient.Update(ctx, clusterConfig)
-	//	if err != nil {
-	//		log.Info(err)
-	//		return
-	//	}
-	//}
+	if (acicontainersoperator.Spec.Flavor >= "openshift-4.3") {
```
snaiksat: What does ">" mean here (since you are comparing strings)? Instead it would be preferable if you can define an array (with currently only one string) which can be later extended to add future openshift versions as we support them. The definition of this should perhaps go into config.go.

pkg/controller/acicontainersoperator.go
```diff
@@ -353,48 +408,57 @@ func (t *OperatorHandler) ObjectCreated(obj interface{}) {
...
+		rclient, err := client.New(cfg, client.Options{Scheme: scheme})
+		if err != nil {
+			return
+		}
+
+		err = rclient.Get(context.TODO(), types.NamespacedName{
+			Name: "cluster",
+		}, clusterConfig)
+		if err != nil {
+			log.Info(err)
```
snaiksat: Okay to proceed in spite of error?

pkg/controller/acicontainersoperator.go
```diff
@@ -353,48 +408,57 @@ func (t *OperatorHandler) ObjectCreated(obj interface{}) {
...
+
+		log.Info("Config Spec is  ", clusterConfig.Spec)
+
+		status  := &configv1.NetworkStatus{}
+
+
+		log.Info("Current clusterConfig.Status is ", clusterConfig.Status)
+
+		if !reflect.DeepEqual(clusterConfig.Status,status){
+			log.Info("Updating Network Status of the network Object....")
```
snaiksat: Can you be more specific (i.e. name of the resource) about the Netowork Object that is being updated here?

Also, why does "Status" start with an upper case? In general, can we exercise a little more discipline when writing log messages at least at the Info level?

pkg/controller/acicontainersoperator.go
```diff
@@ -130,6 +130,49 @@ func NewAciContainersOperator(
...
+}
+
+func (c *Controller) CreateAciContainersOperatorCR() error{
+	log.Info("Reading the Config Map providing CR")
+
+	raw, err := ioutil.ReadFile("/home/config-map")
+
+	log.Debug("CR is ",string(raw))
+
+	//var data interface{}
```
snaiksat: remove if not needed

pkg/controller/acicontainersoperator.go
abhis2112: This is according to pervious suggested comment , and it is like kubernetes for which it fails and will be supported for openshift present/later versions like openshift-4.3/4.4  & so on. No need for an extra array to increase space complexity

pkg/controller/acicontainersoperator.go
abhis2112: This is the passing condition , error handled in earlier leg

pkg/controller/acicontainersoperator.go
snaiksat: I am not very comfortable with the assumption that as far as this flavor value is concerned you are assuming it's either starts with "openshift" or "kubernetes". Moreover this assumption seems to be embedded deep in the code here.

You don't have to use an array, but "space" would hardly be the reason for not using it. ;-)

PR #436, opened 2020-02-20
Fix for ACI operator in reference to Ownereference Object Update

abhis2112:
Issue:
OwnerReference was updated even after the reference was present in aci-containers resources.
Impact:
Aberrant update resulted in only certain resources cleaning up during garbage collection/CR deletion.
Fix:
Check for the presence of the Object before updating it

### DIFF
--- a/pkg/controller/acicontainersoperator.go
+++ b/pkg/controller/acicontainersoperator.go
@@ -300,6 +300,15 @@ func getk8sClient() kubernetes.Interface {
 	return kubeClient
 }
 
+func (t *OperatorHandler) CheckOwnerReference(reference []metav1.OwnerReference) bool{
+	for _,ownerRef := range reference{
+		if ownerRef.Kind == "AciContainersOperator"{
+			log.Debug("OwnerReference Already Present")
+			return true
+		}
+	}
+	return false
+}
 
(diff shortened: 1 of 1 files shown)


pkg/controller/acicontainersoperator.go
```diff
@@ -356,19 +365,21 @@ func (t *OperatorHandler) ObjectCreated(obj interface{}) {
...
-	_, err = deploymentsClient.Update(t.Deployment)
-	if err != nil {
-		log.Error(err.Error())
+	if  !t.CheckOwnerReference(t.Deployment.ObjectMeta.OwnerReferences){
+		t.Deployment.OwnerReferences = []metav1.OwnerReference{
+			*metav1.NewControllerRef(acicontainersoperator, operators.SchemeGroupVersion.WithKind("AciContainersOperator")),
+		}
+		_, err = deploymentsClient.Update(t.Deployment)
+		if err != nil {
+			log.Error(err.Error())
```
snaiksat: Wondering if we should just proceed (here & in other such cases below), or bail out? If the owner reference does not get successfully set, and if the CR is deleted at exactly at this point, we might end up with orphan resources. Not catastrophic, but perhaps something to consider.

PR #439, opened 2020-02-26
Adding support for WebHooks to validate SnatPolicy

jayaramsatya:
1. Added Changes to Validate the SnatIP's and labels across the policies
2. Added some tests validating the Snat policy using Policy raw object included in Admission Request .
TODO:
1. Need to start the HTTP Server to listen  to the Admission request.
2. Need to make the changes to Acc-provision to set up the TLS based communication between the API Server and Acc Controller.

### DIFF
--- a/pkg/controller/webhookserver.go
+++ b/pkg/controller/webhookserver.go
@@ -1,0 +1,230 @@
+// Copyright 2020 Cisco Systems, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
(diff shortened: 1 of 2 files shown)


pkg/controller/webhookserver.go
```diff
@@ -0,0 +1,230 @@
...
+
+	response := &v1beta1.AdmissionReview{}
+	if reviewResponse != nil {
+		response.Response = reviewResponse
+		response.Response.UID = ar.Request.UID
+	}
+	return response, nil
+}
+
+func (cont *AciController) validate(adReview *v1beta1.AdmissionReview) *v1beta1.AdmissionResponse {
```
kishorejonnala: Can we call this as validateCR?

PR #448, opened 2020-03-06
DropLog Event changes

shastrinator:
Add Unix Listener for packet events
DropLogger in opflexagent will send json messages to hostagent.
HostAgent will emit events to src and destination pods for the drops received.

### DIFF
--- a/pkg/hostagent/config.go
+++ b/pkg/hostagent/config.go
@@ -72,6 +72,7 @@ type HostAgentNodeConfig struct {
 
 	// Registry Server URL -- for updating remote EP information
 	RegistryURL string `json:"registry-url,omitempty"`
+
 }
 
 // Configuration for the host agent
@@ -127,6 +128,9 @@ type HostAgentConfig struct {
 
 	// File for writing Opflex server configuration
 	OpFlexServerConfigFile string `json:"opflex-server-config-file,omitempty"`
+
+	// Location of the packet event notification socket which listens to opflex-agent packet events
+	PacketEventNotificationSock string `json:"packet-event-notification-socket,omitempty"`
 
 	// Location of the OVS DB socket
 	OvsDbSock string `json:"ovs-db-sock,omitempty"`
@@ -189,6 +193,21 @@ type HostAgentConfig struct {
 
 	//Namespace for SNAT CRDs
 	AciSnatNamespace string `json:"aci-snat-namespace,omitempty"`
+
+        //DropLogging enabled
+	EnableDropLogging bool `json:"enable-drop-log,omitempty"`
+
+	// DropLog Interface connecting to access bridge
+	DropLogAccessInterface string `json:"drop-log-access-iface,omitempty"`
+
+	// DropLog Interface connecting to access bridge
+	DropLogIntInterface string `json:"drop-log-int-iface,omitempty"`
+
+	// Droplogs older than the expiry-time will be discarded if not published
+	DropLogExpiryTime uint `json:"drop-log-expiry,omitempty"`
+
+	// More than one droplog within the repeat interval for the same event is suppressed
+	DropLogRepeatIntervalTime uint `json:"drop-log-repeat-intvl,omitempty"`
 }
 
 func (config *HostAgentConfig) InitFlags() {
@@ -221,6 +240,7 @@ func (config *HostAgentConfig) InitFlags() {
 	flag.StringVar(&config.OpFlexServerConfigFile, "opflex-server-config-file",
 		"/usr/local/var/lib/opflex-server/config.json",
 		"Config file for opflex server")
+	flag.StringVar(&config.PacketEventNotificationSock, "packet-event-notification-socket", "/usr/local/var/run/aci-containers-packet-event-notification.sock", "Location of the packet event notification socket")
 	flag.StringVar(&config.OvsDbSock, "ovs-db-sock", "/usr/local/var/run/openvswitch/db.sock", "Location of the OVS DB socket")
 	flag.StringVar(&config.EpRpcSock, "ep-rpc-sock", "/usr/local/var/run/aci-containers-ep-rpc.sock", "Location of the endpoint RPC socket used for communicating with the CNI plugin")
 	flag.StringVar(&config.EpRpcSockPerms, "ep-rpc-sock-perms", "", "Permissions to set for endpoint RPC socket file. Octal string")
@@ -247,4 +267,9 @@ func (config *HostAgentConfig) InitFlags() {
 	flag.StringVar(&config.AciVrfTenant, "aci-vrf-tenant", "common", "ACI Tenant containing the ACI VRF for this kubernetes instance")
 	flag.UintVar(&config.Zone, "zone", 8191, "Zone Id for snat flows")
 	flag.StringVar(&config.AciSnatNamespace, "aci-snat-namespace", "aci-containers-system", "Namespace for SNAT CRDs")
+	flag.BoolVar(&config.EnableDropLogging, "enable-drop-log", false, "Allow dropped packets to be logged")
+	flag.StringVar(&config.DropLogAccessInterface, "drop-log-access-iface", "gen2", "Interface in Access bridge to send dropped packets")
+	flag.StringVar(&config.DropLogIntInterface, "drop-log-int-iface", "gen1", "Interface in Integration bridge to send dropped packets")
+	flag.UintVar(&config.DropLogExpiryTime, "drop-log-expiry", 10, "Expiry time for droplogs in the pipeline in minutes")
+	flag.UintVar(&config.DropLogRepeatIntervalTime, "drop-log-repeat-intvl", 2, "Deduplication interval for droplogs of the same event in minutes")
 }
--- a/pkg/hostagent/event_poster.go
+++ b/pkg/hostagent/event_poster.go
@@ -1,0 +1,165 @@
+// Copyright 2020 Cisco Systems, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//	   http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package hostagent
+
+import (
+	v1 "k8s.io/api/core/v1"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/kubernetes/scheme"
+	typedcorev1 "k8s.io/client-go/kubernetes/typed/core/v1"
+	"k8s.io/client-go/tools/record"
+	"k8s.io/client-go/tools/reference"
+	"time"
+	"fmt"
+)
+
+type EventPoster struct {
+	recorder           record.EventRecorder
+	eventSubmitTimeMap map[string]time.Time // (srcIP+dstIP) as key, posted time as value
+}
+
+// Init Poster and return its pointer
+func (agent *HostAgent) initEventPoster(kubeClient *kubernetes.Clientset) {
+	recorder := agent.initEventRecorder(kubeClient)
+	agent.poster = &EventPoster{
+		recorder:           recorder,
+		eventSubmitTimeMap: make(map[string]time.Time),
+	}
+}
+
+// Init Event Recorder for poster object
+func (agent *HostAgent) initEventRecorder(kubeClient *kubernetes.Clientset) record.EventRecorder {
+	eventBroadcaster := record.NewBroadcaster()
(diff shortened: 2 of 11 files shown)


shastrinator:
Hi Jojii, We rate limit the socket 2 ways: Event processing is synchronous, but if we get an event that is late by 10mins(configurable) from current time-> we ignoreit . If we get the same event within 2 minutes(configurable), we ignore it.

pkg/hostagent/event_poster.go
```diff
@@ -0,0 +1,169 @@
...
+	agent.indexMutex.Lock()
+	srcPodKey, srcOk := agent.podIpToName[packetEvent.SourceIP]
+	agent.indexMutex.Unlock()
+	agent.log.Debug("srcPodKey for ", packetEvent.SourceIP, ": ", srcPodKey)
+	if !srcOk {
+		agent.log.Debug(packetEvent.SourceIP, "may not be a pod or not a local pod")
+	} else {
+		obj1, srcExists, err := agent.podInformer.GetStore().GetByKey(srcPodKey)
+		if err == nil {
+			if srcExists {
```
jayaramsatya: can you check obj also here along with srcExists

pkg/hostagent/event_poster.go
```diff
@@ -0,0 +1,169 @@
...
+	}
+	agent.indexMutex.Lock()
+	dstPodKey, dstOk := agent.podIpToName[packetEvent.DestinationIP]
+	agent.indexMutex.Unlock()
+	agent.log.Debug("dstPodKey for ", packetEvent.DestinationIP, ": ", dstPodKey)
+	if !dstOk {
+		agent.log.Debug(packetEvent.SourceIP, "may not be a pod or not a local pod")
+	} else {
+		obj2, dstExists, err := agent.podInformer.GetStore().GetByKey(dstPodKey)
+		if err == nil {
```
jayaramsatya: same here please check both obj2, dstExists

pkg/hostagent/event_poster.go
```diff
@@ -0,0 +1,169 @@
...
+							packetEvent.DropReason); err != nil {
+						return err
+					}
+				}
+			}
+		} else {
+			srcExists = false
+		}
+		srcOk = srcExists
+	}
```
jojimt: As discussed, let's post one drop event to only one of the pods (i.e. in the corner case where both src and dest are local, pick one and post it there, not to both). Also, please account for the NAT from the agent side if possible. Otherwise, we would end up missing lots of events.

pkg/hostagent/packet_event.go
```diff
@@ -0,0 +1,89 @@
...
+                    fd.Close()
+                    return
+                }
+                var m [30]PacketEvent
+                err1 := json.Unmarshal(buffer[:n], &m)
+                agent.log.Infof("PacketEvent Received length %d \n",n)
+                if err1 != nil {
+			agent.log.Info("Unmarshaling error ", err1)
+                }
+		numRecords := n/int(reflect.TypeOf(m[0]).Size())
```
jayaramsatya: nit: just run go fit

jojimt: This would be a good place to get time.Now()

pkg/hostagent/event_poster.go
shastrinator: Sure.

pkg/hostagent/packet_event.go
shastrinator: Sure will do

pkg/hostagent/packet_event.go
```diff
@@ -0,0 +1,89 @@
...
+                for i := 0; i < numRecords; i++ {
```
jojimt: Please use for _, event := range m. You don't need to compute the size.

pkg/hostagent/packet_event.go
```diff
@@ -0,0 +1,89 @@
...
+                buffer := make([]byte, 4096)
+                n, err := fd.Read(buffer)
+                if err != nil {
+                    if err != io.EOF {
+                        agent.log.Errorf("packet event socket read error %s\n", err)
+                    }
+                    fd.Close()
+                    return
+                }
+                var m [30]PacketEvent
```
jojimt: Don't specify a size here. Unmarshal will allocate it.

pkg/hostagent/packet_event.go
```diff
@@ -0,0 +1,89 @@
...
+			agent.log.Info("Unmarshaling error ", err1)
+                }
+		numRecords := n/int(reflect.TypeOf(m[0]).Size())
+                for i := 0; i < numRecords; i++ {
+			err2 := agent.processPacketEvent(m[i])
+			if err2 != nil {
+				agent.log.Errorf("Failed to post event %d\n", err2)
+			}
+                }
+            } ()
```
jayaramsatya: fd needs to be closed ? before existing from go routine

pkg/hostagent/packet_event.go
```diff
@@ -0,0 +1,89 @@
...
+        }
+    } ()
+
+    go func() {
+        <-stopCh
+        pc.Close()
+    } ()
+    return
```
jojimt: please remove the redundant return

pkg/hostagent/packet_event.go
shastrinator: Thanks I will add a defer

pkg/hostagent/event_poster.go
shastrinator: sure

pkg/hostagent/config.go
```diff
@@ -247,4 +267,9 @@ func (config *HostAgentConfig) InitFlags() {
 	flag.StringVar(&config.AciVrfTenant, "aci-vrf-tenant", "common", "ACI Tenant containing the ACI VRF for this kubernetes instance")
 	flag.UintVar(&config.Zone, "zone", 8191, "Zone Id for snat flows")
 	flag.StringVar(&config.AciSnatNamespace, "aci-snat-namespace", "aci-containers-system", "Namespace for SNAT CRDs")
+	flag.BoolVar(&config.EnableDropLogging, "enable-drop-log", true, "Allow dropped packets to be logged")
+	flag.StringVar(&config.DropLogAccessInterface, "drop-log-access-iface", "gen2", "Interface in Access bridge to send dropped packets")
+	flag.StringVar(&config.DropLogIntInterface, "drop-log-int-iface", "gen1", "Interface in Integration bridge to send dropped packets")
+	flag.UintVar(&config.DropLogExpiryTime, "drop-log-expiry", 10, "Expiry time for droplogs in the pipeline")
+	flag.UintVar(&config.DropLogRepeatIntervalTime, "drop-log-repeat-intvl", 2, "Deduplication interval for droplogs of the same event")
```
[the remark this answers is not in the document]
shastrinator: Sure will do.

pkg/hostagent/event_poster.go
```diff
@@ -0,0 +1,169 @@
...
+// Submit an event using kube API with message attached
+func (agent *HostAgent) submitEvent(pod *v1.Pod, message string, dropReason string) error {
+	agent.log.Info("Posting event ", message)
+	ref, err := reference.GetReference(scheme.Scheme, pod)
+	if err != nil {
+		agent.log.Error("Returning ", err)
+		return err
+	}
+	if dropReason == "" {
+		dropReason = "Dropped"
```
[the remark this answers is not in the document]
shastrinator: Ok.

pkg/hostagent/event_poster.go
shastrinator: Sure I will think about the NAT case more. I will limit the drop event to src Pod in case both are present.

pkg/hostagent/packet_event.go
shastrinator: ok

pkg/hostagent/packet_event.go
shastrinator: ok

pkg/hostagent/packet_event.go
shastrinator: ok thanks

pkg/hostagent/event_poster.go
```diff
@@ -0,0 +1,169 @@
...
+func (agent *HostAgent) shouldIgnore(packetEvent PacketEvent) bool {
+	// ignore if the given packetDrop is out of date
+	logTime, err := time.Parse(time.UnixDate, packetEvent.TimeStamp)
+	currTime := time.Now()
+	if (err != nil) || (currTime.Sub(logTime).Minutes() > float64(agent.config.DropLogExpiryTime)) {
+		agent.log.Debug("Ignoring expired event")
+		return true
+	}
+	// ignore if the event has been posted within the defined time period
+	key := packetEvent.SourceIP + packetEvent.DestinationIP
```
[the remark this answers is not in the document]
shastrinator: This is rate limiting for the event in general, not specific to pod. But in general we do want to capture all the different events for a pod

pkg/hostagent/event_poster.go
```diff
@@ -0,0 +1,169 @@
...
+	}
+	if agent.shouldIgnore(packetEvent) {
+		return nil
+	}
+	agent.indexMutex.Lock()
+	srcPodKey, srcOk := agent.podIpToName[packetEvent.SourceIP]
+	agent.indexMutex.Unlock()
+	agent.log.Debug("srcPodKey for ", packetEvent.SourceIP, ": ", srcPodKey)
+	if !srcOk {
+		agent.log.Debug(packetEvent.SourceIP, "may not be a pod or not a local pod")
```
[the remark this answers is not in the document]
shastrinator: They are not guaranteed to be pod Ips. I will have to handle the NAT case as you mention. Will think about it.

mchalla:
> * Yes, you can create the droplog interfaces and enable/disable on the fly using a watched config file( packet-drop-log-file-config path in opflex-agent). Besides there is a DropLogEnable config as part of the agent config- that is set to true by default, but can be turned off.

Ok. my question is where do you delete them if the feature is turned off ?

shastrinator:
> > * Yes, you can create the droplog interfaces and enable/disable on the fly using a watched config file( packet-drop-log-file-config path in opflex-agent). Besides there is a DropLogEnable config as part of the agent config- that is set to true by default, but can be turned off.
> 
> Ok. my question is where do you delete them if the feature is turned off ?
I was assuming since we don't ignore these ports in the case the feature is turned off, they will naturally get deleted. I will test these again.

pkg/hostagent/packet_event.go
shastrinator: Sure. Am passing it as an argument now

pkg/hostagent/packet_event.go
shastrinator: Thanks Done

pkg/hostagent/packet_event.go
shastrinator: ok.

pkg/hostagent/packet_event.go
shastrinator: Done thanks

pkg/hostagent/event_poster.go
shastrinator: Yes I have to address NAT still. I aam posting only for the source pod now

pkg/hostagent/event_poster.go
```diff
@@ -0,0 +1,169 @@
...
+		agent.poster.recorder.Event(ref, v1.EventTypeWarning, dropReason, message)
+	}
+	return nil
+}
+
+// Check if given PacketEvent should be ignored
+func (agent *HostAgent) shouldIgnore(packetEvent PacketEvent) bool {
+	// ignore if the given packetDrop is out of date
+	logTime, err := time.Parse(time.UnixDate, packetEvent.TimeStamp)
+	currTime := time.Now()
```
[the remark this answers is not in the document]
shastrinator: Yes Done

pkg/hostagent/event_poster.go
```diff
@@ -0,0 +1,169 @@
...
+	eventBroadcaster.StartRecordingToSink(
+		&typedcorev1.EventSinkImpl{Interface: kubeClient.CoreV1().Events("")})
+	component := "aci-containers-host"
+	recorder := eventBroadcaster.NewRecorder(scheme.Scheme, v1.EventSource{Component: component})
+	return recorder
+}
+
+// Submit an event using kube API with message attached
+func (agent *HostAgent) submitEvent(pod *v1.Pod, message string, dropReason string) error {
+	agent.log.Info("Posting event ", message)
```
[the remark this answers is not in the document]
shastrinator: We will not even know whether the agent is posting events, if we are in info level. I thought this will be good to verify.

jayaramsatya:
Hi Kiran, 
Please ignore pod updates for the event changes as we are adding update role in cluster, it means when you add event you will get PodUpdate that needs to be ignored, And also before merging the change need to discuss in adding cluster role update for User Pods.

shastrinator:
Hi Jayaram,
I am not using podUpdates for anything other than indexing Ip to name, so
nothing to ignore there.Sure, we should discuss the pod update role changes
-Kiran

shastrinator:
Added unit test cases

PR #449, opened 2020-03-07
Added Changes to handle the Port Exhaustion when we apply SnatPolicy

jayaramsatya:
1. Added Changes to handle the Port Exhaustion, ones ports and IP's gets exhausted will set the policy status to IpPortsExhausted as shown in below 
kubectl describe snatpolicy it-ns-ns-snatpolicy
Metadata:
  Creation Timestamp:  2020-04-08T21:46:25Z
  Generation:          1
  Resource Version:    4743194
  Self Link:           /apis/aci.snat/v1/snatpolicies/it-ns-ns-snatpolicy
  UID:                 5acebb94-61ac-48e9-b985-afe6d06e86c8
Spec:
  Dest Ip:
    100.100.100.0/24
  Snat Ip:
    10.10.0.8
Status:
  State:  IpPortsExhausted
Events:   <none> 

2. Generated the Snatpolicy client code with nonNamespace role,  which is to support policy at cluster level 

3. Added Config map watcher for to handle Snat-operator-config Configmap Updates
   Currently ConfigMap is created from aci deployment  and values for ConfigMap shown below,   with new code changes by adding watcher for modifications to this ConfigMap(snat-operator-config),  user can edit this ConfigMap with different values. 
apiVersion: v1
kind: ConfigMap
metadata:
  name: snat-operator-config
  namespace: aci-containers-system
  labels:
    aci-containers-config-version: "340ded98-6fb0-4b06-a542-f30c4bfeb8c8"
    network-plugin: aci-containers
data:
    "start": "5000"
    "end": "65000"
    "ports-per-node": "3000"

4. Currently there is no support to handle delete and addition of Config maps. as we are expecting config map creation is driven from aci deployment. if user deletes the config map 
until ACI controller restarts most recent config will be used.  once after restart it will fallback to default value  ports-per-node 3000


snaiksat:
@jayaramsatya  Thanks, but as discussed, could you please update the commit message (not just the PR description) indicating that you are adding the config-map watcher. More importantly, please clearly provide an example from a user's perspective as to how they would update the specific config-map value.

PR #455, opened 2020-03-11
Temporarily removing veth_host interface and EP file - part 2

snaiksat:
Since we are seeing some asymmetrical routing on account
of the route to the pod network added to support the
use of the veth_host interface.

### DIFF
--- a/pkg/hostagent/nodes.go
+++ b/pkg/hostagent/nodes.go
@@ -18,12 +18,7 @@ package hostagent
 
 import (
 	"encoding/json"
-	"fmt"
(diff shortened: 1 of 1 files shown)


jojimt:
Removing this only for on-prem, right?

snaiksat:
@jojimt this is being done on a jordan-eft branch which is not being used for overlay, only for EFT on on-prem.

jojimt:
oh ok

snaiksat:
Tests need to be updated but merging this for now since we need an urgent build.

PR #460, opened 2020-03-15
Added support to  add extra service file for openshift-ingress service

jayaramsatya:
Added support to  add extra service file for openshift-ingress service
This service file contains OpenShift Infrastrucre IngressIP  for Service-ip mappings

### DIFF
--- a/pkg/hostagent/services.go
+++ b/pkg/hostagent/services.go
@@ -15,6 +15,7 @@
 package hostagent
 
 import (
+	"context"
 	"encoding/json"
 	"fmt"
 	"io/ioutil"
@@ -24,15 +25,17 @@ import (
 	"strings"
 
 	"github.com/Sirupsen/logrus"
-
+	configv1 "github.com/openshift/api/config/v1"
 	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/watch"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/tools/cache"
-
 	"k8s.io/kubernetes/pkg/controller"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/client/config"
 )
 
 type opflexServiceMapping struct {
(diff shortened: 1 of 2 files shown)


pkg/hostagent/services.go
```diff
@@ -63,6 +66,28 @@ type opflexService struct {
...
+
+// Namespace of Openshift Service
+const (
+	OpenShiftIngress string = "openshift-ingress"
+)
+
+// Represent the Openshift services
+type opflexOcService struct {
+	Name      string
+	Namespace string
```
snaiksat: Why isn't there an IP filed here? See my other comments below.

pkg/hostagent/services.go
```diff
@@ -326,8 +350,41 @@ func (agent *HostAgent) updateServiceDesc(external bool, as *v1.Service,
...
 			agent.opflexServices[ofas.Uuid] = ofas
-			return true
+			if agent.config.AciVmmDomainType == "OpenShift" {
+				if !external {
+					for _, v := range agent.ocServices {
+						if v.Name != as.ObjectMeta.Name &&
+							v.Namespace != as.ObjectMeta.Namespace {
+							continue
+						}
+						var ingressIp string
```
snaiksat: This processing is specific to just the ingressIp. We need the ability to be able to add other IPs as well. The plan was to have a list, each element of which would be a 3-tuple: (namespace, service-name, ip-addres)

That way we can populate the list with all the known the desired 3-tuples and generically process them here. Now I think you will need to special processing for every service we add.

pkg/hostagent/services.go
```diff
@@ -63,6 +66,28 @@ type opflexService struct {
...
+
+// Name of the Openshift Service
+const (
+	RouterInternalDefult string = "router-internal-default"
+)
+
+// Namespace of Openshift Service
+const (
+	OpenShiftIngress string = "openshift-ingress"
+)
```
snaiksat: This initialization should be on a list of struct (which will support the ability to add more).

pkg/hostagent/services.go
```diff
@@ -436,3 +499,44 @@ func (agent *HostAgent) updateAllServices() {
...
+	infraStructureInfo := &configv1.Infrastructure{
+		TypeMeta:   metav1.TypeMeta{APIVersion: configv1.GroupVersion.String(), Kind: "Infrastructure"},
+		ObjectMeta: metav1.ObjectMeta{Name: "cluster"},
+	}
+	cfg, err := config.GetConfig()
+	scheme := runtime.NewScheme()
+	scheme.AddKnownTypes(configv1.SchemeGroupVersion, &configv1.Infrastructure{})
+	err = configv1.Install(scheme)
+	if err != nil {
+		agent.log.Error("failed to Install: ", err)
```
snaiksat: As a follow up lets put a more relevant message.

pkg/hostagent/services.go
```diff
@@ -436,3 +499,44 @@ func (agent *HostAgent) updateAllServices() {
...
+		return ""
+	}
+	rclient, err := client.New(cfg, client.Options{Scheme: scheme})
+	if err != nil {
+		return ""
```
snaiksat: lets log

pkg/hostagent/services.go
```diff
@@ -436,3 +499,44 @@ func (agent *HostAgent) updateAllServices() {
...
+	}
+	if rclient == nil {
+		return ""
```
snaiksat: lets log

PR #461, opened 2020-03-17
Added extra serivce files for the services ApiServer and default-dns

jayaramsatya:
This Service Files contains OpenShift Infrastrucre DnsIP and API Server Internal IP  will be used as service ip mapping

### DIFF
--- a/pkg/hostagent/services.go
+++ b/pkg/hostagent/services.go
@@ -66,20 +66,18 @@ type opflexService struct {
 	Attributes map[string]string `json:"attributes,omitempty"`
 }
 
-// type of Openshift Infrastructure Ip's
-const (
-	IngressIp string = "IngressIp"
-	DnsIp     string = "DnsIp"
-)
-
 // Name of the Openshift Service
 const (
-	RouterInternalDefult string = "router-internal-default"
+	RouterInternalDefault string = "router-internal-default"
+	DnsDefault            string = "dns-default"
+	ApiServer             string = "kubernetes"
 )
 
 // Namespace of Openshift Service
 const (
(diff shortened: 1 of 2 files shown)


pkg/hostagent/services.go
```diff
@@ -66,20 +66,18 @@ type opflexService struct {
...
+	DnsDefault           string = "dns-default"
+	ApiServer            string = "kubernetes"
 )
 
 // Namespace of Openshift Service
 const (
-	OpenShiftIngress string = "openshift-ingress"
+	OpenShiftIngressNs string = "openshift-ingress"
+	OpenShiftDnsNs            = "openshift-dns"
+	DefaultNs          string = "defaullt"
```
snaiksat: There is a typo in the "default" namespace name.

pkg/hostagent/services.go
```diff
@@ -522,19 +513,19 @@ func (agent *HostAgent) getInfrastucreIp(ipType string) string {
...
 		return ""
 	}
 	if infraStructureInfo.Status.Platform == configv1.OpenStackPlatformType {
 		if infraStructureInfo.Status.PlatformStatus != nil &&
 			infraStructureInfo.Status.PlatformStatus.OpenStack != nil {
-			agent.log.Debug("IngressIP: ", infraStructureInfo.Status.PlatformStatus.OpenStack.IngressIP)
-			switch ipType {
-			case IngressIp:
+			switch serviceName {
+			case RouterInternalDefult:
```
snaiksat: There is a typo in this name, but I think its consistently wrong so I guess we might be okay!

PR #463, opened 2020-03-26
Fixed the issue to  omit network address and broadcast address

jayaramsatya:
When user provides SnatIP  lets say 10.10.10.1/24 the first IP selected as 10.10.10.0, 
 which is network address,  now to fix this issue we are omitting  network address and broadcast address


snaiksat:
Merged in master, if required we can backport from there.

PR #473, opened 2020-04-03
Make host access configurable

mchalla:
- enabled for overlay mode
- disabled for on prem


mchalla:
verified make check-hostagent

PR #480, opened 2020-04-07
Move OVS/OpFlex containers to alpine 3.11.5

tomflynn:
Remove doxygen from opflex-build-base
Move to python3 for OVS and apply required patch for 2.12 (fixed in next release)


tomflynn:
OK I pushed another PR to update that as well

PR #482, opened 2020-04-08
Fixed the test case TestServiceGraph

jayaramsatya:
It Is the  test case issue from day one, it is working occasionally because before delete is completely processed node adds are trigged so still stale data is present. In this test as node is annotated with IP 10.6.1.1,   when node gets deleted the IP is added back to IPAM. And when the node comes back IPAM allocates the same IP from start,  So as node annotations are same, Service sync is not trigged to populate APIC data to be written.  now test case is fixed by setting node annotations with different values

### DIFF
--- a/pkg/controller/services_test.go
+++ b/pkg/controller/services_test.go
@@ -268,14 +268,14 @@ func TestServiceAnnotation(t *testing.T) {
 	}
 	twoNodeRedirect := redirect(seMap{
 		"node1": &metadata.ServiceEndpoint{
-			HealthGroupDn:  "uni/tn-common/svcCont/redirectHealthGroup-kube_svc_node1",
-			Mac:			"8a:35:a1:a6:e4:60",
-			Ipv4:			net.ParseIP("10.6.1.1"),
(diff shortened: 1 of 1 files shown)


pkg/controller/services_test.go
```diff
@@ -628,6 +628,11 @@ func TestServiceGraph(t *testing.T) {
 	cont.fakeNodeSource.Delete(node1)
 	cont.fakeNodeSource.Delete(node2)
 	sgWait(t, "delete node", cont, map[string]apicapi.ApicSlice{name: nil})
+	//reset the node annotations with diffrent values before adding the node
+	node1.ObjectMeta.Annotations[metadata.ServiceEpAnnotation] =
+		"{\"mac\":\"8a:35:a1:a6:e4:60\",\"ipv4\":\"10.6.1.5\"}"
+	node2.ObjectMeta.Annotations[metadata.ServiceEpAnnotation] =
+		"{\"mac\":\"a2:7e:45:57:a0:d4\",\"ipv4\":\"10.6.1.6\"}"
```
jojimt, line 635: @jayaramsatya can we clear the field instead to mimic a real use case?

PR #492, opened 2020-04-13
Migrating from dep to gomod

jayaramsatya:


jojimt:
@jayaramsatya please make two separate commits, one for vendor changes and another or non-vendor code. Otherwise, we won't be able to view the diffs.

PR #497, opened 2020-04-15
Migrating from dep to gomod

jayaramsatya:
1. Added newer version cni spec 0.4.0
2. Make changes are done to remove building vendor
3. Upgraded the kubernetes client package to version 1.17


cmd/opflexagentcni/main.go
```diff
@@ -285,9 +285,45 @@ func cmdDel(args *skel.CmdArgs) error {
...
+			return err
+		}
+	}
+	if n.NetConf.RawPrevResult == nil {
+		return fmt.Errorf("Required prevResult missing")
+	}
+	if err := version.ParsePrevResult(&n.NetConf); err != nil {
+		return err
+	}
+	result, err := current.NewResultFromResult(n.PrevResult)
```
kishorejonnala, line 312: Thanks for taking care of this PrevResult part.

.travis.yml
```diff
@@ -3,9 +3,8 @@ matrix:
   include:
     - language: go
       sudo: false
-      go: "1.10"
+      go: "1.13"
```
snaiksat, line 6: We need to confirm that the base container image with which we ship (i.e. UBI) supports 1.13.

go.mod
```diff
@@ -0,0 +1,130 @@
...
+	k8s.io/api v0.17.3
+	k8s.io/apimachinery v0.17.3
+	k8s.io/client-go v8.0.0+incompatible
+	k8s.io/kubernetes v0.0.0-00010101000000-000000000000
+	sigs.k8s.io/controller-runtime v0.5.0
+)
+
+replace (
+	go.etcd.io/bbolt => go.etcd.io/bbolt v1.3.3
+	k8s.io/api => k8s.io/kubernetes/staging/src/k8s.io/api v0.0.0-20200211180626-06ad960bfd03
```
snaiksat, line 85: After doing this, and some of the other version pining below, to what extent will we be compatible with an older version of kubernetes API/server? There will be users who will have deployed k8s < 1.17, either directly, or by virtue of using a distro like OpenShift/RH, and we wont be able to force them to upgrade their k8s deployment.

jayaramsatya: Sumit, it should be backward compatible, may be we should validate, if we have any 1.13 older version fabs

.travis.yml
jayaramsatya: Go modules full support introduced from go 1.12 on-words.

go.mod
jayaramsatya: BTW Currently we are using 1.15 API version

.travis.yml
snaiksat: I am not referring to gomod, I am referring to the golang version 1.13, and verifying it being available in the UBI base container version that we intend to use. For instance, RHEL 7 did not have golang 1 .13.

go.mod
snaiksat: Why just 1.13? Elsewhere below you have version pinned to 1.17. So what about any version < 1.17? We will have fewer, perhaps if any customers on, say 1.13, but we will definitely customers on 1.16 (also need to confirm what OpenShift 3.11 has).

PR #499, opened 2020-04-20
Added service type and nodeport to service file

jayaramsatya:


snaiksat:
As a follow up, can you please add UTs which exercises each of these types?

PR #501, opened 2020-04-21
Fix incorrect table in nftables programming

shastrinator:
Use nat instead of ip filter as the table


shastrinator:
Have tested the functionality with nftables locally.

PR #503, opened 2020-04-23
Added ability to watch Aci Deployment/Daemonset to update the OwnerReference of Aci Operator

abhis2112:
The fix correctness can be tested in the following ways:
1. Deleting the CR and observing all Aci resources terminates.
2. Delete any deployment/daemonset in aci-containers-system namespaces and adding it back again to check if owner reference is added.
3. Check the network status is added by the Operator or not.
4. Delete the operator pod itself and check whether it reconcile to the same state.
5. Creating CR manually to check if ACI resources are spawned or not

### DIFF
--- a/pkg/controller/acicontainersoperator.go
+++ b/pkg/controller/acicontainersoperator.go
@@ -32,7 +32,6 @@ import (
 	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/apimachinery/pkg/watch"
 	"k8s.io/client-go/kubernetes"
-	restclient "k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/cache"
 	"k8s.io/client-go/util/workqueue"
 	"os"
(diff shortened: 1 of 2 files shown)


jojimt:
@abhis2112 please run gofmt.

pkg/controller/acicontainersoperator.go
```diff
@@ -97,35 +99,116 @@ func NewAciContainersOperator(
...
+			key, err := cache.MetaNamespaceKeyFunc(obj)
+			log.Debug("Added Deployment key: %s", key)
+			if err == nil {
+				deployment_queue.Add(key)
+			}
+		},
+		UpdateFunc: func(prevObj, currentObj interface{}) {
+			key, err := cache.MetaNamespaceKeyFunc(currentObj)
+			log.Debug("Updated Deployment: %s", key)
+			if err == nil {
```
jayaramsatya, line 179: do you want to check here if there is any change in prevObj and currentObj

pkg/controller/acicontainersoperator.go
```diff
@@ -311,124 +389,134 @@ func (t *OperatorHandler) CheckOwnerReference(reference []metav1.OwnerReference)
...
+	log.Debug(acicontainersoperator.Spec.Config)
+
+	if (acicontainersoperator.Spec.Config == ""){
+		log.Info("acicnioperator CR config is Nil")
+		return true
+	}
+	
+	dec, err := base64.StdEncoding.DecodeString(acicontainersoperator.Spec.Config)
+	if err != nil {
+		panic(err)
```
jayaramsatya, line 492: if you want to retry , i guess you need to remove panic. applicable every where you have put panic

pkg/controller/acicontainersoperator.go
```diff
@@ -311,124 +389,134 @@ func (t *OperatorHandler) CheckOwnerReference(reference []metav1.OwnerReference)
...
+		return true
+	}
+	log.Info("Applying Aci Deployment")
+
+	//Currently the Kubectl version is v.1.14. This will be updated by the acc-provision according
+	//to the platform specification
+	cmd := exec.Command("kubectl","apply","-f","aci-deployment.yaml")
+	log.Debug(cmd)
+	_, _ = cmd.Output()
```
jayaramsatya, line 519: Probably it is better to do retry here if you are applying cmd, check for error and do retry

pkg/controller/acicontainersoperator.go
```diff
@@ -481,17 +569,88 @@ func (t *OperatorHandler) ObjectCreated(obj interface{}) {
...
-				return
+				return true
 			}
 		}
 	}
+
+	log.Info("Adding Aci Operator OwnerRefrence to resources ....")
+	c.indexMutex.Lock()
+	if (!(c.UpdateDeploymentOwnerReference(acicontainersoperator))){
+		log.Info("Error Updating Deployment Owner Reference")
```
jayaramsatya, line 580: This will be problem if you return from here without unlock, may be it is better move your lock inside the function so that you no need to take care every where

pkg/controller/acicontainersoperator.go
```diff
@@ -206,85 +284,85 @@ func (c *Controller) Run(stopCh <-chan struct{}) {
...
+			case string:
+				obj, exists, err := store.GetByKey(key)
+				if err == nil && exists {
+					log.Info("Controller.processNextItem: object Creation detected:", key)
+					requeue = createhandler(obj)
+				}
+
+				if !exists {
+					log.Info("Controller.processNextItem: object deleted detected:", key)
+					deletehandler(key)
```
jayaramsatya, line 351: do you want to handle later your delete handler ? as you are not doing anything in the dunction

abhis2112: Currently, yes it is only logging information about whether the object is deleted or not. I am using it as an endpoint if in future we want anything in deletehandler

pkg/controller/acicontainersoperator.go
abhis2112: I will check for the type of error. Sometimes it's a generic message where things are applied but still, it shows a message like "please apply your changes to the latest version and try again". I will do a fix where we can handle for any other error conditions

pkg/controller/acicontainersoperator.go
abhis2112: Noted .. handling it

pkg/controller/acicontainersoperator.go
abhis2112: Noted.. providing check

pkg/controller/acicontainersoperator.go
```diff
@@ -97,101 +97,207 @@ func NewAciContainersOperator(
...
+		Resources:           AciResources{},
 	}
 
 	log.Info("Adding Event Handlers")
-	informer.AddEventHandler(cache.ResourceEventHandlerFuncs{
+	aci_operator_informer.AddEventHandler(cache.ResourceEventHandlerFuncs{
 		AddFunc: func(obj interface{}) {
 			key, err := cache.MetaNamespaceKeyFunc(obj)
-			log.Infof("The acioperator key: %s", key)
+			log.Debug("Added acicontainersoperator  key:", key)
```
jayaramsatya, line 145: nit just give space after key:

pkg/controller/acicontainersoperator.go
```diff
@@ -206,233 +315,240 @@ func (c *Controller) Run(stopCh <-chan struct{}) {
...
-		}
-
-		_, err = ovsdaemonsetclient.Update(t.OvsDaemonset)
-		if err != nil {
-			log.Error(err.Error())
-		}
+	_, err = cmd.Output()
+	if err != nil {
+		log.Error(err)
+		return false
```
jayaramsatya, line 546: here you need to return true ?

PR #505, opened 2020-04-27
Add policing to drop-log interfaces

shastrinator:
To prevent excessive dropped packets from swamping the CPU,
add 1 Mbps policing to drop-log interface. Packet size can be 12Kb
at the most(MTU).But typically around 250Bytes. This allows for
around 500 pps for 250B packets or around 83 pps for 1500B.
Minor fix for issue caused by another commit.

### DIFF
--- a/pkg/hostagent/ovs.go
+++ b/pkg/hostagent/ovs.go
@@ -253,7 +253,25 @@ func (agent *HostAgent) diffPorts(bridges map[string]ovsBridge) []libovsdb.Opera
 				}
 				ops = append(ops, adds...)
 			}
-
+                       agent.ignoreOvsPorts[agent.config.IntBridgeName] = []string{agent.config.DropLogIntInterface}
+                       accbr, ok := bridges[agent.config.AccessBridgeName]
+                       if ok {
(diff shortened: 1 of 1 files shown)


pkg/hostagent/ovs.go
```diff
@@ -276,26 +294,6 @@ func (agent *HostAgent) diffPorts(bridges map[string]ovsBridge) []libovsdb.Opera
 			}
 			agent.ignoreOvsPorts[agent.config.IntBridgeName] = []string{"pi-veth_host_ac"}
 			agent.ignoreOvsPorts[agent.config.AccessBridgeName] = []string{"pa-veth_host_ac"}
-			if agent.config.EnableDropLogging {
```
mchalla, line 279: Why are we removing this conditional ? Isnt it nice to have it as a config option ?

shastrinator: It is not removed. The block that it is moved to has the same condition

shastrinator: and the reason it was moved was because overlay condition was added to the existing block.

PR #506, opened 2020-04-27
Update opflex containers to use UBI 8 instead of alpine

tomflynn:
This changeset updates the base image for our opflex/ovs containers to be UBI 8

Unlike alpine, devel packages for libuv and rapidjson are not available so the build now builds/installs them manually to allow compilation to proceed


snaiksat:
Which image tag should we be using this to test this?

tomflynn:
You can use the the ubi-builds tag to test with.
openvswitch, opflex and opflex-server containers have been updated

mandeepdhami:
+1

PR #508, opened 2020-04-28
Istio-1.5: Moving from 1.4 to upstream version 1.5

kishorejonnala:
+ Istio-1.5 addressed scale/performance issues and significantly different from 1.4
+ Snapshot of istioctl (version 1.5.2) is added to our repo for pinning.
+ Auto create istio-system NS as part of istiocontrolplane CR

### DIFF
--- a/docker/Dockerfile-controller
+++ b/docker/Dockerfile-controller
@@ -3,10 +3,10 @@ RUN apk upgrade --no-cache
 RUN apk update && apk add curl git
 RUN curl -LO https://storage.googleapis.com/kubernetes-release/release/`curl -s https://storage.googleapis.com/kubernetes-release/release/stable.txt`/bin/linux/amd64/kubectl
 RUN chmod u+x kubectl && mv kubectl /usr/local/bin/kubectl
-RUN curl -LO https://istio.io/operator.yaml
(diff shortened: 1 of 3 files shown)


docker/Dockerfile-controller
```diff
@@ -5,8 +5,8 @@ RUN curl -LO https://storage.googleapis.com/kubernetes-release/release/`curl -s
 RUN chmod u+x kubectl && mv kubectl /usr/local/bin/kubectl
 RUN curl -LO https://istio.io/operator.yaml
 RUN mkdir -p /usr/local/var/lib/aci-cni
-COPY pkg/istiocrd/upstream-istio-operator-1.4.0.yaml /usr/local/var/lib/aci-cni/upstream-istio-operator.yaml
-COPY pkg/istiocrd/upstream-istio-cr-1.4.0.yaml /usr/local/var/lib/aci-cni/upstream-istio-ctrlplane-resource.yaml
+COPY pkg/istiocrd/istioctl-1.5.2 /usr/local/bin/istioctl
```
snaiksat, line 8: Is it an option to download istioctl like what we are doing with kubectl?

kishorejonnala: @snaiksat good point; apparently there is a way and I fixed it.

PR #514, opened 2020-05-01
Update dockerfile to use ubi for gbpserver and the dev environment

jojimt:
- Created dev dockerfiles for hostagent and controller, using noirolabs/ubibase as the base. This allows building without a RHEL env. 
- Switched gbpserver to use ubi.
- Updated Makefile to use dev dockerfiles if Mac env is detected.


jojimt:
@tomflynn

PR #523, opened 2020-05-16
Change order of veth patch ports and ep files

mchalla:
- currently launch script creates ep file
- go code creates patch ports
- this can cause issues when agent sees the ep file
  before the patch ports are completely setup
- make the launch code create ep.gen file
- go code creates patch ports
- copies the .gen file into the endpoints dir
- in addition the nodes.go reads the file at init
- make it read the .gen file if ep file is missing

### DIFF
--- a/pkg/hostagent/pods.go
+++ b/pkg/hostagent/pods.go
@@ -292,9 +292,14 @@ func (agent *HostAgent) syncEps() bool {
 	seen := make(map[string]bool)
 	nullMacFile := false
 	nullMacCheck := agent.getEpFileName(agent.config.DefaultEg.Name)
+        vethHostFile := false
 	for _, f := range files {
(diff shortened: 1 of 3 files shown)


pkg/hostagent/pods.go
```diff
@@ -409,6 +417,26 @@ func (agent *HostAgent) creatNullMacEp() {
 
 }
 
+func (agent *HostAgent) creatVethHostEp() {
```
jojimt, line 420: Would be better add this check agent.config.OpflexMode == "overlay" and exit. That way, this will execute only if the mode is overlay, even if some files are present.

snaiksat:
I believe this was already merged in https://github.com/noironetworks/aci-containers/pull/524

mchalla:
Yes Sumit.

PR #526, opened 2020-05-17
Route informer is enabled conditional

jayaramsatya:
1. Enabled the Route informer if the openshift apiserver is running
2. Openshift API server creates the routes. so it is ideal to enable the route informer
  once api sever is running.
3. Fixed the issue when updating DnsOperator Spec if the nodeaddress list is nil update fails.
   so avoided the update of the spec

### DIFF
--- a/pkg/controller/acicontainersoperator.go
+++ b/pkg/controller/acicontainersoperator.go
@@ -35,6 +35,7 @@ import (
 	log "github.com/sirupsen/logrus"
 	appsv1 "k8s.io/api/apps/v1"
 	v1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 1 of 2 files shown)


pkg/controller/acicontainersoperator.go
```diff
@@ -397,9 +399,14 @@ func (c *Controller) CreateAciContainersOperatorCR() error {
...
 		if er != nil {
-			log.Info(er)
-			log.Info("Waiting for CRD to get registered to etcd....")
-			return false, nil
+			if errors.IsAlreadyExists(er) { //Happens due to etcd timeout
+				log.Info(er)
+				return true, nil
+			} else {
+				log.Info(er)
+				log.Info("Wssaiting for CRD to get registered to etcd....")
```
snaiksat, line 407: nit: Typo, but since its info message, better to correct

PR #527, opened 2020-05-18
Updating dockerfiles to include lables/licenses for Openshift Container Images

abhis2112:


docker/Dockerfile-gbpserver
```diff
@@ -1,5 +1,13 @@
 FROM registry.access.redhat.com/ubi8/ubi:latest
 COPY dist-static/gbpserver /usr/local/bin/
-
+# Required OpenShift Labels
+LABEL name="ACI CNI gbserver" \
```
jojimt, line 4: Pls change "gbserver" to "gbpserver" everywhere in this file.

docker/Dockerfile-operator
```diff
@@ -1,7 +1,16 @@
 FROM registry.access.redhat.com/ubi8/ubi:latest
-RUN yum --disablerepo=\*ubi\* install -y curl git \
-  && yum clean all \
-  && curl -LO https://storage.googleapis.com/kubernetes-release/release/v1.14.6/bin/linux/amd64/kubectl \
-  && chmod u+x kubectl && mv kubectl /usr/local/bin/kubectl
+#RUN yum --disablerepo=\*ubi\* install -y curl git \
```
jojimt, line 2: Is this commenting intentional?

abhis2112: checking....

docker/Dockerfile-gbpserver
abhis2112: sure...

snaiksat:
Thanks @abhis2112 . Since some of the information here is public facing (via OpenShift operator hub) we should revisit and polish it a little more in a follow up patch.

PR #531, opened 2020-05-21
Checking container status of Openshift-apiserver container

abhis2112:
Adding provision to start Route node informer when openshift-apiserverr container is in Ready State

### DIFF
--- a/pkg/controller/acicontainersoperator.go
+++ b/pkg/controller/acicontainersoperator.go
@@ -1071,7 +1071,7 @@ func (c *Controller) enableRouteInformer(stopCh <-chan struct{}) {
 		var options metav1.ListOptions
 		for {
 			Pods, err := c.K8s_Clientset.CoreV1().Pods("openshift-apiserver").List(options)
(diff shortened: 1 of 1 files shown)


pkg/controller/acicontainersoperator.go
```diff
@@ -1071,7 +1071,7 @@ func (c *Controller) enableRouteInformer(stopCh <-chan struct{}) {
 		var options metav1.ListOptions
 		for {
 			Pods, err := c.K8s_Clientset.CoreV1().Pods("openshift-apiserver").List(options)
-			if err == nil && (len(Pods.Items) > 0 && Pods.Items[0].Status.Phase == v1.PodRunning) {
+			if err == nil && (len(Pods.Items) > 0 && Pods.Items[0].Status.ContainerStatuses[0].Ready == true) {
```
jayaramsatya, line 1074: do you want to check for both the container ID's ?

snaiksat: this might perhaps not be necessary..picking this change for now, we can decide as follow up whether we need to check for both containers

PR #536, opened 2020-05-27
Fixed the crash seen in snat UT.the crash is due to data race condition.

jayaramsatya:
Did small reorg of the code


kishorejonnala:
LGTM

PR #542, opened 2020-05-29
Adding ACI Operator Minor Changes

abhis2112:
1. New Map for DNS Operator changes so to limit its usage to OpenStack platform.
2. Added cloud flavor to version-map in aci-operator to support AWS.
3. Removing return, if CR not created by the operator and keep running informers.
4. Updating the status of CR to false if spec.config is nil.

### DIFF
--- a/pkg/controller/acicontainersoperator.go
+++ b/pkg/controller/acicontainersoperator.go
@@ -84,8 +84,11 @@ type Controller struct {
 
 var Version = map[string]bool{
 	"openshift-4.3": true,
+	"cloud": true,
 }
-
+var Dnsoper = map[string]bool{
+	"openshift-4.3": true,
+}
 const aciContainersController = "aci-containers-controller"
 const aciContainersHostDaemonset = "aci-containers-host"
 const aciContainersOvsDaemonset = "aci-containers-openvswitch"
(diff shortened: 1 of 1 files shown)


pkg/controller/acicontainersoperator.go
```diff
@@ -639,6 +642,11 @@ func (c *Controller) handleOperatorCreate(obj interface{}) bool {
 
 	if acicontainersoperator.Spec.Config == "" {
 		log.Info("acicnioperator CR config is Nil...Exiting")
+		acicontainersoperator.Status.Status = false
+		_, er := c.Operator_Clientset.AciV1alpha1().AciContainersOperators(os.Getenv("SYSTEM_NAMESPACE")).Update(acicontainersoperator)
```
jayaramsatya, line 646: Did you verify? that Status is getting updated.  there is call yo update only status UpdateStatus.

abhis2112: yes verified on snatjordan cluster , with CR as nil and status updated as false
apiVersion: aci.ctrl/v1alpha1
kind: AciContainersOperator
metadata:
  creationTimestamp: "2020-05-29T22:34:24Z"
  generation: 2
  name: acicnioperator
  namespace: aci-containers-system
  resourceVersion: "929236"
  selfLink: /apis/aci.ctrl/v1alpha1/namespaces/aci-containers-system/acicontainersoperators/acicnioperator
  uid: 9b3437c5-1d87-4e5c-9aff-22d061ead9e0
spec:
  config: ""
  flavor: ""
status:
  Successful: false

PR #543, opened 2020-06-04
aci-containers  CF/PCF code artifacts is removed. since it is not supported

jayaramsatya:


kishorejonnala:
@jayaramsatya Can you trigger a regression with these changes to make sure there are no dependencies on opflex side?
(Ex: https://github.com/noironetworks/opflex/pull/2)

PR #546, opened 2020-06-09
Remove unneeded libs from opflex_agent/opflex_server containers

tomflynn:
Reduce the size of a couple of containers

opflex_agent container doesn't need grpc libs
opflex_server container doesn't need OVS libs

Removing them saves ~42MB for the agent and ~11MB for the server.

TODO: Additional savings could be had by fixing the symlinks so that we don't have multiple copies of each lib in /usr/local/lib

10.30.120.20/noiro/opflex-server                tom-test           a308727b78e9   About an hour ago   306 MB
10.30.120.20/noiro/opflex                       tom-test           19ed391bdc3a   About an hour ago   276 MB
10.30.120.20/noiro/opflex-server                master-test        9839f9df5c22   45 minutes ago      317 MB
10.30.120.20/noiro/opflex                       master-test        2e13dd41c9bd   About an hour ago   318 MB

### DIFF
--- a/docker/Dockerfile-opflex
+++ b/docker/Dockerfile-opflex
@@ -16,8 +16,8 @@ COPY licenses /licenses
 COPY bin/opflex_agent /usr/local/bin/
 COPY bin/mcast_daemon /usr/local/bin/
 COPY bin/gbp_inspect /usr/local/bin/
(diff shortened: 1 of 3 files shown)


tomflynn:
Converting PR to draft until I have time to test the containers

PR #553, opened 2020-06-27
Added Endpointslice support

jayaramsatya:
1. Added watchers to listen for Endpoint slices changes both in controller and hostagent code
2. Service points to multiple endpointslice Object and each Slice Object points to list of Endpoints
   made changes for the Services code accordingly.
3. Egress netwokpolicy changes are done to handle the Endpoint Slices
4. Currently endpointslice support flag is turned off for this commit
5. Added multiple  UT's in services and networkpolicies by enabling the endpointslice support flag in tests

### DIFF
--- a/pkg/controller/network_policy.go
+++ b/pkg/controller/network_policy.go
@@ -41,6 +41,7 @@ import (
 	"github.com/noironetworks/aci-containers/pkg/index"
 	"github.com/noironetworks/aci-containers/pkg/ipam"
 	"github.com/noironetworks/aci-containers/pkg/util"
+	v1beta "k8s.io/api/discovery/v1beta1"
 )
 
 func (cont *AciController) initNetworkPolicyInformerFromClient(
@@ -694,6 +695,21 @@ func checkEndpoints(subnetIndex cidranger.Ranger,
 
 	return true
 }
+func checkEndpointslices(subnetIndex cidranger.Ranger,
+	addresses []string) bool {
+
+	for _, addr := range addresses {
+		ip := net.ParseIP(addr)
+		if ip == nil {
+			return false
+		}
+		contains, err := subnetIndex.Contains(ip)
+		if err != nil || !contains {
+			return false
+		}
+	}
+	return true
+}
 
 type portRemoteSubnet struct {
 	port           *v1net.NetworkPolicyPort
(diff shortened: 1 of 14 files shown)


pkg/controller/network_policy.go
```diff
@@ -795,71 +811,138 @@ func (cont *AciController) getServiceAugmentBySubnet(subj apicapi.ApicObject,
...
+							proto, port, service)
+
+						logger.WithFields(logrus.Fields{
+							"proto":   proto,
+							"port":    port,
+							"service": servicekey,
+						}).Debug("Allowing egress for service by subnet match")
+					}
+				})
+		}
```
kishorejonnala, line 943: Can we move this code into a function? Possibly collapse both cases (endpoint slice and endpoint) into one routine.

kishorejonnala: @jojimt  Pls take a look fron Cloud perspective (ex: populating Region/Zone information)

jayaramsatya: Sure Kishore will take a look

pkg/controller/network_policy.go
```diff
@@ -802,62 +818,116 @@ func (cont *AciController) getServiceAugmentBySubnet(subj apicapi.ApicObject,
...
 		service := serviceobj.(*v1.Service)
-		endpoints := endpointsobj.(*v1.Endpoints)
-
-		for _, svcPort := range service.Spec.Ports {
-			if prs.port != nil &&
-				(svcPort.Protocol != *prs.port.Protocol ||
-					svcPort.TargetPort.String() !=
-						prs.port.Port.String()) {
-				// egress rule does not match service target port
+		if !cont.endPointSliceEnabled {
```
jojimt, line 825: Can we avoid this check everywhere in the code? We should define a go interface or function pointer that is setup at init and does the appropriate thing.

pkg/controller/services.go
```diff
@@ -978,18 +1056,50 @@ func (cont *AciController) opflexDeviceDeleted(dn string) {
...
-			key + ": " + err.Error())
-		return
-	}
-
 	aobj := apicapi.NewVmmInjectedSvc(cont.vmmDomainProvider(),
 		cont.config.AciVmmDomain, cont.config.AciVmmController,
 		service.Namespace, service.Name)
 	aobjDn := aobj.GetDn()
 	aobj.SetAttr("guid", string(service.UID))
+	if !cont.endPointSliceEnabled {
```
jojimt, line 1064: Let's avoid this if check and use some abstraction so this code doesn't need to know the details of the k8s object.

pkg/hostagent/agent.go
```diff
@@ -138,6 +140,7 @@ func NewHostAgent(config *HostAgentConfig, env Environment, log *logrus.Logger)
 				DefaultNs,
 			},
 		},
+		endPointSliceEnabled: false,
```
jojimt, line 143: This should come from config, right? You can default the config to false.

jayaramsatya: Yes Joji, i will change this code reading it from config

pkg/controller/network_policy.go
jayaramsatya: Sure will create interface class for all this functions

jayaramsatya:
LGTM  ACC provision changes are still pending once that is merged. we can merge this code.

jayaramsatya:
Addressed some abstraction at the acc controller code.  still some pending code in hostagent will take care of it in next commit.

PR #560, opened 2020-07-10
CSCvu92591: Handling DeletedFinalStateUnknown

kishorejonnala:
K8S resourceEventHandler can return a DeletedFinalStateUnknown object in some scenarios, in which case we parse it accordingly and handle without panicing.


snaiksat:
As a follow up optimization patch we could perhaps factor the code into a common function which takes a resource type as a parameter and processes it accordingly?

kishorejonnala:
@snaiksat  thought about it - it's not going to be straightforward given different resource types use different APIs - the code will still be bunch of case handling - So I thought this would be more readable.

PR #579, opened 2020-07-30
Use filter-chain for SNAT PBR contract

kishorejonnala:
The changes to address the issue https://github.com/noironetworks/support/issues/1213


jayaramsatya:
Hi Kishore, could you please check because of indentation issue, some of the old diffs also shown in the current diff.

PR #583, opened 2020-07-31
 Clean Up the Stale metadata and files.

jayaramsatya:
Stale data can be  possible if we miss any register events for unconfiguring the container interface
 Ideally epMetadata for the given pod  should contain only one ConatinerID entry
 As every pod contains one network namespace for all the containers with in the pod
 (this  containerID -> pause container serves as parent container)
 remove the  containerid from epmetadata which doesn't match the Container ID passed by the
 configure interface call.

### DIFF
--- a/pkg/hostagent/setup.go
+++ b/pkg/hostagent/setup.go
@@ -292,6 +292,37 @@ func (agent *HostAgent) configureContainerIfaces(metadata *md.ContainerMetadata)
 			return nil, err
 		}
 	}
+	var StaleContMetadata []md.ContainerMetadata
+	podid := metadata.Id.Namespace + "/" + metadata.Id.Pod
+	agent.indexMutex.Lock()
+	if len(agent.epMetadata[podid]) > 1 {
+		logger.Warnf("There is a Stale metadata present for the pod")
+	}
+	if val, ok := agent.epMetadata[podid]; ok {
+		// check for any stale entry present.
+		// this is possible if we miss any register events for unconfiguring the POD
+		//  ideally epMetadata[podid] Map should contain only one entry of containerID
+		// As every pod contains one network namespace for all the containers with in the pod
+		// (i.e pause container ==> metadata.Id.ContId)
+		// if there are any stale which doesn't match the ContainerID remove it
+		for key, v := range val {
+			if metadata.Id.ContId != key {
+				logger.Warnf("Stale metadata present clean the entry: %s", key)
+				StaleContMetadata = append(StaleContMetadata, *v)
+			}
+		}
+	}
+	agent.indexMutex.Unlock()
+
+	for _, v := range StaleContMetadata {
+		err := agent.cleanStatleMetadata(v.Id.ContId)
+		if err == nil {
+			agent.deallocateMdIps(&v)
+			agent.ipamMutex.Lock()
+			delete(agent.epMetadata[podid], v.Id.ContId)
+			agent.ipamMutex.Unlock()
+		}
+	}
 
 	err := md.RecordMetadata(agent.config.CniMetadataDir,
 		agent.config.CniNetwork, *metadata)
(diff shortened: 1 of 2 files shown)


pkg/hostagent/setup.go
```diff
@@ -301,22 +301,53 @@ func (agent *HostAgent) configureContainerIfaces(metadata *md.ContainerMetadata)
...
-				make(map[string]*md.ContainerMetadata)
+	agent.indexMutex.Lock()
+	if val, ok := agent.epMetadata[podid]; ok {
+		// check for any stale entry present.
+		// this is possible if we miss any register events for unconfiguring the POD
+		//  ideally epMetadata[podid] Map should contain only one entry of containerID
+		// As every pod contains one network namespace for all the containers with in the pod
+		// (i.e pause container ==> metadata.Id.ContId)
+		// if there are any stale which doesn't match the ContainerID remove it
+		// @TODO This logic needs to be revisited in case if pod contains multiple network namespaces
```
jojimt, line 312: I think this TODO can be removed. Single n/w namespace per pod is fundamental to k8s. No point in keeping a TODO for something so basic.

pkg/hostagent/setup.go
```diff
@@ -301,22 +301,53 @@ func (agent *HostAgent) configureContainerIfaces(metadata *md.ContainerMetadata)
...
+				}
+			}
 		}
-		agent.epMetadata[podid][metadata.Id.ContId] = metadata
-		agent.indexMutex.Unlock()
+	} else {
+		agent.epMetadata[podid] =
+			make(map[string]*md.ContainerMetadata)
 	}
+	agent.epMetadata[podid][metadata.Id.ContId] = metadata
```
jojimt, line 329: Let's add a length check and logger.Warnf if len(epMetadata[podid] > 1).

pkg/hostagent/setup.go
```diff
@@ -301,22 +301,53 @@ func (agent *HostAgent) configureContainerIfaces(metadata *md.ContainerMetadata)
...
+		// this is possible if we miss any register events for unconfiguring the POD
+		//  ideally epMetadata[podid] Map should contain only one entry of containerID
+		// As every pod contains one network namespace for all the containers with in the pod
+		// (i.e pause container ==> metadata.Id.ContId)
+		// if there are any stale which doesn't match the ContainerID remove it
+		// @TODO This logic needs to be revisited in case if pod contains multiple network namespaces
+		for key, v := range val {
+			if metadata.Id.ContId != key {
+				logger.Info("Stale ContainerId entry: ", key)
+				err := agent.cleanStatleMetadata(key)
```
jojimt, line 316: We should avoid doing this under lock as we are accessing the file system. Instead, can we generate a local list and do the delete after we release the lock?

pkg/hostagent/setup.go
```diff
@@ -301,22 +301,53 @@ func (agent *HostAgent) configureContainerIfaces(metadata *md.ContainerMetadata)
 	}
 
 	podid := metadata.Id.Namespace + "/" + metadata.Id.Pod
-	{
-		agent.indexMutex.Lock()
-		if _, ok := agent.epMetadata[podid]; !ok {
-			agent.epMetadata[podid] =
-				make(map[string]*md.ContainerMetadata)
+	agent.indexMutex.Lock()
```
jojimt, line 304: Please move this section to be before the RecordMetadata. This way, we will ensure delete-before-create.

pkg/hostagent/setup.go
jayaramsatya: Ok

pkg/hostagent/setup.go
jayaramsatya: noted will make the change

pkg/hostagent/setup.go
jayaramsatya: ok

pkg/hostagent/setup.go
```diff
@@ -292,6 +292,32 @@ func (agent *HostAgent) configureContainerIfaces(metadata *md.ContainerMetadata)
...
+			}
+		}
+	}
+	agent.indexMutex.Unlock()
+
+	for _, v := range StaleContMetadata {
+		err := agent.cleanStatleMetadata(v.Id.ContId)
+		if err == nil {
+			agent.deallocateMdIps(&v)
+			delete(agent.epMetadata[podid], v.Id.ContId)
```
jojimt, line 318: Please move this delete from the map to be inside the lock (line 309 is fine).

PR #591, opened 2020-08-06
Matching the pods label with snatlabel for other namespaces.

jayaramsatya:
When we apply the Snatpolicy having lable with namespace. if existing pod with matching snatlabel
in any other namespace exists for this pod snatpolicy is applied.
This is problem is there only for pod resource. others are fine.
Moved the code to common function to avoid this issue.


snaiksat:
Perhaps its difficult to test this in a UT?

PR #596, opened 2020-08-10
CSCvv23505: Changing SGT connection's adjType to L3

kishorejonnala:
Issue: In cases where ACI CNI created SG is consumed by multiple EPGs spreaded across different BDs, adjType L2 results in a fault on APIC.

Fix: Change the adjType of SGT's vnsAbsConnection to L3 so that ACI CNI created SG can be usable in more deployment scenarios.


kishorejonnala:
Travis run is failing; I remember this particular test "TestServiceGraph" was flaky on Indus and it got fixed later on.

PR #597, opened 2020-08-11
CSCvu76204: Istio resource clean-up with ACI-CNI undeploy

kishorejonnala:
**Issue:** When aci-cni deployment file is deleted, some of the resources such as istio-operator NS are not deleted.

AciIstioOperator CRD presence in aci-cni deployment drives istio installation which includes
+ istio-operator NS
+ istio-operator deployment
+ istio-system NS
+ istiocontrolplane CR

**Fix:** This change will set the OwnerReference with AciIstioOperator for all resources that gets created by ACC.

Caveat: There are bunch of istio related CRDs that gets created by upstream istio-operator deployment are not cleaned up even after istio-operator deployment goes away. This is a known issue in upstream community (https://github.com/istio/istio/issues/24038)


pkg/controller/aci_istio.go
```diff
@@ -145,6 +149,58 @@ func (cont *AciController) handleIstioUpdate(istiospec *istiov1.AciIstioOperator
...
 	}
+
+	cfg, err := config.GetConfig()
+	iop := &iopv1alpha1.IstioOperator{}
+	scheme := runtime.NewScheme()
+	iopSchemeVersion := schema.GroupVersion{Group: "install.istio.io", Version: "v1alpha1"}
+	scheme.AddKnownTypes(iopSchemeVersion, &metav1.GetOptions{}, &metav1.UpdateOptions{}, iop)
+	istioclient, err := client.New(cfg, client.Options{Scheme: scheme})
+	if err != nil {
+		return true
```
jayaramsatya, line 160: Nit can you log error here

pkg/controller/aci_istio.go
```diff
@@ -145,6 +149,58 @@ func (cont *AciController) handleIstioUpdate(istiospec *istiov1.AciIstioOperator
...
+	}
+
+	istioNamespaces := [...]string{"istio-system", "istio-operator"}
+	nsClient := kubeClient.CoreV1().Namespaces()
+	if nsClient == nil {
+		cont.log.Error("Error in Fetching nsClient")
+		return true // to requeue
+	}
+	for _, ns := range istioNamespaces {
+		istioNs, _ := nsClient.Get(context.TODO(), ns, metav1.GetOptions{})
```
jayaramsatya, line 188: can you use the local cache to get the Namespace Object cont.namespaceIndexer.GetByKey(<nsname>) this way you can use local cache

jayaramsatya: Key you can pass namespace name

kishorejonnala: Good point; fixed.

PR #598, opened 2020-08-12
Port Exhaustion issue is fixed

jayaramsatya:
Edit the SnatOperator config map and make sure that snatPorts get exhausted.
and then bring down the portrange, here snatpolicy state should change back From IpPortsExhausted to Ready state,
which is not happening.
To fix this issue snatpolicy state is recomputed  when ever snatnodeinfo or snat configmap changes, by reading the snatpolicy state from snatpoilcy Indexstore
also caching of the PortsExhausted state is removed as it is not required with new changes

### DIFF
--- a/pkg/controller/snatglobalinfo.go
+++ b/pkg/controller/snatglobalinfo.go
@@ -118,17 +118,21 @@ func (cont *AciController) snatCfgUpdate(obj interface{}) {
 		start < 5000 || end > 65000 || start > end || portsPerNode > end-start+1 {
 		return
 	}
-	cont.indexMutex.Lock()
 	portRange.Start = start
 	portRange.End = end
 	var currPortRange []snatglobalinfo.PortRange
 	currPortRange = append(currPortRange, portRange)
+	var nodeInfoKeys []string
+	cont.indexMutex.Lock()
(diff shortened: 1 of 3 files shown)


pkg/controller/snatglobalinfo.go
```diff
@@ -433,49 +464,28 @@ func (cont *AciController) clearSnatGlobalCache(policyName string, nodename stri
...
 			cont.log.Debug("handleSnatPoilcyUpdate Queu the Key: ")
-			cont.queueNodeInfoUpdateByKey(nodeinfokey)
+			nodeinfokeys = append(nodeinfokeys, (nodeinfokey))
 		}
 	}
+	return nodeinfokeys
 }
 
-func (cont *AciController) handleSnatPolicyPortAllocFailures(allocfailed map[string]bool, nodename string) bool {
+func (cont *AciController) SetSnatPoliciesState(names map[string]bool, status snatv1.PolicyState) bool {
```
kishorejonnala, line 482: Nit: Fix the naming convention

kishorejonnala: LGTM; You can follow-up with a UT if possible.

PR #608, opened 2020-08-19
Check added for snat Port-range boundry condition

jayaramsatya:
As previous commit removed the caching of port-exhaustion state in controller.
now every time hostgent sees the change in snatnode info because of snatpolicy state(IpPortExhasted)it has the push the updates
so now that condition is not required.


pkg/controller/snatglobalinfo_test.go
```diff
@@ -247,5 +247,15 @@ func TestSnatCfgChangeTest(t *testing.T) {
 		"node-1": {SnatIp: "10.1.1.9", PortRanges: []snatglobalinfo.PortRange{{Start: 10000, End: 14999}}},
 	}
 	snatWait(t, "snat test", expected, cont.AciController.snatGlobalInfoCache["10.1.1.9"])
```
snaiksat, line 249: for checking boundary conditions do you need to add tests for (1) end is the same as start, (2) end is one more than start, (3) end is less than start ?

jayaramsatya: Yeah i have given this range in the config map "start": "5000", "end": "65000", "ports-per-node": "60000" so that start+ports-per-node >= 65000

PR #616, opened 2020-09-01
Update cni-deploy-container to avoid CVE

shastrinator:
https://github.com/rancher/rancher/issues/27369
Thanks for the info Tom.


tomflynn:
The dep in the hostagent code needs to be updated as well so we pick up the fix for SetupVeth

PR #621, opened 2020-09-03
Move to OVS 2.13

tomflynn:


shastrinator (approved):
Alpine opflex-build-base/openvswitch will have to be updated to 2.13 as well I guess.

tomflynn:
> Alpine opflex-build-base/openvswitch will have to be updated to 2.13 as well I guess.

alpine doesn't give us a way to choose specific OVS versions. They have 2.12 for now but I imagine edge will be moving to 2.13 pretty soon

PR #625, opened 2020-09-08
Dynamic detection of CRDs

jojimt:
- Added infra for dynamically detecting CRDs. This means the corresponding handler code is only invoked if the CRD is detected
 - Moved qos handling code to this new mode as reference

### DIFF
--- a/pkg/controller/crds.go
+++ b/pkg/controller/crds.go
@@ -46,6 +46,7 @@ func (cont *AciController) initCRDInformerBase(listWatch *cache.ListWatch) {
 				crd, ok := obj.(*v1.CustomResourceDefinition)
 				if !ok {
 					cont.log.Errorf("Add CRD bad object")
+					return
 				}
 				cont.crdAdded(crd)
 			},
--- a/pkg/controller/crds.go
+++ b/pkg/controller/crds.go
@@ -1,0 +1,71 @@
+// Copyright 2020 Cisco Systems, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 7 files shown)


pkg/controller/crds.go
```diff
@@ -0,0 +1,71 @@
...
+}
+
+func (cont *AciController) initCRDInformerBase(listWatch *cache.ListWatch) {
+	_, cont.crdInformer = cache.NewInformer(
+		listWatch, &v1.CustomResourceDefinition{}, 0,
+		cache.ResourceEventHandlerFuncs{
+			AddFunc: func(obj interface{}) {
+				crd, ok := obj.(*v1.CustomResourceDefinition)
+				if !ok {
+					cont.log.Errorf("Add CRD bad object")
```
jayaramsatya, line 48: can we return here ? as we are accessing  the object in crdAdded

pkg/controller/qos.go
```diff
@@ -47,6 +51,20 @@ func QosPolicyLogger(log *logrus.Logger, qos *qospolicy.QosPolicy) *logrus.Entry
...
+func qosInit(cont *AciController, stopCh <-chan struct{}) {
+	cont.log.Debug("Initializing qos client")
+	restconfig := cont.env.RESTConfig()
+	qosClient, err := qosclientset.NewForConfig(restconfig)
+	if err != nil {
+		cont.log.Errorf("Failed to intialize qos client")
+		return
+	}
+	cont.initQosInformerFromClient(qosClient)
+	cache.WaitForCacheSync(stopCh, cont.qosInformer.HasSynced)
```
jayaramsatya, line 63: can we move the informer run before the waitForCache Sync

jojimt: I think I will eliminate the WaitForCacheSync. It doesn't serve any purpose in this workflow actually.

pkg/controller/crds.go
jojimt: Yes, good catch.

jayaramsatya (approved):
LGTM

PR #629, opened 2020-09-17
Added Support for Namedport based networkpolicy

jayaramsatya:
Namedport is instead of numeric port, it gives another layer of indirection and maps to diffrent numeric ports for an endpoint.
example multiple http servers can expose target ports as 80 or 8080.
now in that case all egress direction  services matching the network policy needs to match all the ports matching the name.
Added UT's to cover the above cases. And verifed all the upstream testcases

### DIFF
--- a/pkg/controller/network_policy.go
+++ b/pkg/controller/network_policy.go
@@ -36,6 +36,7 @@ import (
 	"k8s.io/apimachinery/pkg/util/intstr"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/tools/cache"
+	k8util "k8s.io/kubectl/pkg/util"
 
 	"github.com/noironetworks/aci-containers/pkg/apicapi"
 	"github.com/noironetworks/aci-containers/pkg/index"
@@ -510,6 +511,8 @@ func (cont *AciController) updateTargetPortIndex(service bool, key string,
 				networkPolicyKeys: make(map[string]bool),
 			}
 			cont.targetPortIndex[portkey] = entry
+		} else {
+			entry.port.ports = port.ports
 		}
 
 		if service {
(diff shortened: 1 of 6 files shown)


pkg/controller/network_policy.go
```diff
@@ -519,26 +522,78 @@ func (cont *AciController) updateTargetPortIndex(service bool, key string,
...
 }
+func (cont *AciController) getPortNumsFromPortName(podKeys []string, portName string) []int {
+	var ports []int
+	portmap := make(map[int]bool)
+	for _, podkey := range podKeys {
+		podobj, exists, err := cont.podIndexer.GetByKey(podkey)
+		if exists && err == nil {
+			pod := podobj.(*v1.Pod)
+			port, err := k8util.LookupContainerPortNumberByName(*pod, portName)
+			if err != nil {
```
kishorejonnala, line 533: Can you log the error here.

pkg/controller/network_policy.go
```diff
@@ -519,26 +522,78 @@ func (cont *AciController) updateTargetPortIndex(service bool, key string,
...
+				continue
+			}
+			if _, ok := portmap[int(port)]; !ok {
+				ports = append(ports, int(port))
+				portmap[int(port)] = true
```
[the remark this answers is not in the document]
jayaramsatya: Added the logs  here to print PortName to Port numbers

pkg/controller/pods.go
```diff
@@ -94,14 +95,50 @@ func (cont *AciController) handlePodUpdate(pod *v1.Pod) bool {
...
+				portmap, ok := cont.ctrPortNameMap[ctrportspec.Name]
+				if !ok {
+					cont.ctrPortNameMap[ctrportspec.Name] = make(map[string]uint)
+					cont.ctrPortNameMap[ctrportspec.Name][key]++
+					nmport = true
+				} else {
+					if _, found := portmap[key]; !found {
+						nmport = true
+					}
+					portmap[key]++
```
jojimt, line 117: I think this refcount won't be reliable. You might need to keep a reverse map.

pkg/controller/pods.go
```diff
@@ -202,7 +239,37 @@ func (cont *AciController) podDeleted(obj interface{}) {
...
 	cont.removePodFromNode(pod.Spec.NodeName, podkey)
+	nmport := false
+	for _, ctr := range pod.Spec.Containers {
+		for _, ctrportspec := range ctr.Ports {
+			if ctrportspec.Name != "" {
+				portmap, ok := cont.ctrPortNameMap[ctrportspec.Name]
+				if ok {
+					key := portProto(&ctrportspec.Protocol) + "-" + strconv.Itoa(int(ctrportspec.ContainerPort))
+					if portmap[key] == 1 {
+						delete(cont.ctrPortNameMap, ctrportspec.Name)
```
jojimt, line 250: Could you get a port name change as part of a pod update as well?

pkg/controller/pods.go
jayaramsatya: Ok will maintain the reverse map

pkg/controller/pods.go
jayaramsatya: no, i am not able to edit the port name in pod spec

snaiksat (approved):

PR #630, opened 2020-09-22
Set default veth MTU based on infra VLAN uplink

tomflynn:
The value can be explicitly set using interface-mtu


jayaramsatya (approved):
LGTM

PR #632, opened 2020-09-24
Fixed the crash seen while writing the service file

jayaramsatya:
This is a Data race issue becuase of  concurrent map iteration and map write.
Reason for this issue is Opflex pointer is not deep copied.
And the same pointer is beeing used while writing the file.
it is observed in services and ep sync functions.

### DIFF
--- a/pkg/hostagent/pods.go
+++ b/pkg/hostagent/pods.go
@@ -41,6 +41,7 @@ import (
 	"k8s.io/kubernetes/pkg/controller"
 
 	"github.com/noironetworks/aci-containers/pkg/metadata"
+	"github.com/noironetworks/aci-containers/pkg/util"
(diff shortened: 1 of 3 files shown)


kishorejonnala (dismissed):
LGTM

pkg/hostagent/agent.go
```diff
@@ -394,3 +394,14 @@ func (agent *HostAgent) Run(stopCh <-chan struct{}) {
 
 	agent.cleanupSetup()
 }
+func deepCopyObj(src, dst interface{}) error {
```
snaiksat, line 397: would it make sense to put this in util package?

jayaramsatya: moved to util

snaiksat (approved):

PR #637, opened 2020-09-29
Recreate operator CR on operator start/restart w.r.t changes in configmap

abhis2112:
The sequence of Operation:
1. New aci_deployment.yaml files when run acc-provision for new images/configuration
2. Apply aci_deployment.yaml
3. New aci-operator-configmap will be applied.
4. Current CR Spec.config and configmap Spec.config is compared if different new CR will be deployed
   with respective changes in configmap


jayaramsatya (approved):
LGTM

PR #646, opened 2020-10-14
Scale down of Node leaving the Stale Snat nodeInfoCR  entry

jayaramsatya:
after bringing down the nodes, snat Nodeinfo CR is not cleaned up.
Now added a explicit call to delete the Nodeinfo upon the delete of Node

### DIFF
--- a/pkg/controller/snatglobalinfo.go
+++ b/pkg/controller/snatglobalinfo.go
@@ -231,6 +231,12 @@ func (cont *AciController) handleSnatNodeInfo(nodeinfo *nodeinfo.NodeInfo) bool 
 		ret = cont.deleteNodeinfoFromGlInfoCache(nodename)
 		updated = true
 	} else {
+		// This case ignores any stale entry is present in nodeinfo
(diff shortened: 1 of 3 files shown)


pkg/controller/snatglobalinfo.go
```diff
@@ -231,6 +231,12 @@ func (cont *AciController) handleSnatNodeInfo(nodeinfo *nodeinfo.NodeInfo) bool
 		ret = cont.deleteNodeinfoFromGlInfoCache(nodename)
 		updated = true
 	} else {
+		// This case ignores any stale entry is present in nodeinfo
+		_, _, err := cont.nodeIndexer.GetByKey(nodename)
+		if err != nil {
+			cont.log.Error("Could not lookup node: ", err, "nodeName: ", nodename)
```
snaiksat, line 237: hate to nit pick, but this should not be an "error" level, right? this should perhaps just be at "info" level?

jayaramsatya: Yeah good point

snaiksat (approved):

PR #651, opened 2020-10-19
Added flavor for OSP 4.5 Openstack/ESX in ACI Operator

abhis2112:


snaiksat (approved):

snaiksat:
We are not supporting 4.5 in release branch so this will not be required

PR #654, opened 2020-10-20
Creating open-shift service file copies for openshift-Esx

jayaramsatya:
Service file copy created for openshift ingress using installer provisioned lb IP


abhis2112 (approved):

snaiksat:
Thanks @jayaramsatya . As a follow up, at the very least, I think it will help to make the logging more specific for all the cases, so that we know for which service which VIP IP is used (this is the comment for the OpenShift on OpenStack processing) and when the IP is being read from the config-map (in case of ESX) versus from infra CRD (in case of OpenShift).

PR #655, opened 2020-10-20
Creating open-shift service file copies for openshift-Esx

jayaramsatya:
Service file copy created for openshift ingress using installer provisioned lb IP

(cherry picked from commit 7e5b23b80086f223d966162f98b059b7abab6121)

### DIFF
--- a/pkg/hostagent/services.go
+++ b/pkg/hostagent/services.go
@@ -89,6 +89,11 @@ type opflexOcService struct {
 	Namespace string
 }
 
+var Version = map[string]bool{
+	"openshift-4.4-esx": true,
+	"openshift-4.5-esx": true,
+}
+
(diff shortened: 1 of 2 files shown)


abhis2112 (approved):

pkg/hostagent/services.go
```diff
@@ -524,6 +529,12 @@ func (agent *HostAgent) updateAllServices() {
 
 // This API is get the OpenShift InfrastructreIp's
 func (agent *HostAgent) getInfrastucreIp(serviceName string) string {
+	if _, ok := Version[agent.config.Flavor]; ok {
+		if serviceName == RouterInternalDefault {
```
snaiksat, line 533: It would have desirable to post a warning log if this was an empty string.

PR #656, opened 2020-10-20
Add support to compile iptables from sources

mchalla:
- For ubi containers we do not have a iptables-legacy package
- To work around this compile iptables from sources in base image
- Copy iptables artifacts from base image into dist-static
  (done in jenkins scripts)
- Dockerfile-host will pick the artifacts and install it locally
  in its container
- Install iptables alternatives to
  - /usr/sbin/iptables-legacy
  - /usr/sbin/iptables-nft
- Run wrapper that will set initial alternative to
  - /usr/sbin/iptables-wrapper
- At run time depending on which rules are used by kubelet
  the alternative will be modfied to use iptables-legacy
  or iptables-nft

Signed-off-by: Madhu Challa <challa@gmail.com>

### DIFF
--- a/docker/Dockerfile-opflex-build-base
+++ b/docker/Dockerfile-opflex-build-base
@@ -5,13 +5,30 @@ RUN microdnf install --enablerepo codeready-builder-for-rhel-8-x86_64-rpms \
     libtool pkgconfig autoconf automake make cmake file python2-six \
     openssl-devel git gcc gcc-c++ boost-devel diffutils python2-devel \
     libnetfilter_conntrack-devel wget which curl-devel procps zlib-devel \
+    libmnl-devel vi \
     && microdnf clean all
-RUN wget https://github.com/Tencent/rapidjson/archive/v1.1.0.tar.gz \
+RUN git clone -b libnftnl-1.1.7 https://git.netfilter.org/libnftnl \
+  && cd libnftnl \
+  && ./autogen.sh \
+  && ./configure \
+  && make $make_args \
+  && make install && cp libnftnl.pc /usr/lib64/pkgconfig && make clean \
+  && cd / \
+  && rm -Rf libnftnl \
+  && git clone -b v1.8.5 https://git.netfilter.org/iptables \
+  && cd iptables \
+  && ./autogen.sh \
+  && ./configure --prefix=/usr --sbindir=/sbin \
+  && make $make_args \
+  && make install && make clean \
+  && cd / \
+  && rm -Rf iptables \
+  && wget https://github.com/Tencent/rapidjson/archive/v1.1.0.tar.gz \
   && tar xvfz v1.1.0.tar.gz \
   && cd rapidjson-1.1.0 \
   && cmake CMakeLists.txt \
   && cp -R include/rapidjson/ /usr/local/include/ \
-  && mkdir /usr/local/lib/pkgconfig \
+  && mkdir -p /usr/local/lib/pkgconfig \
   && cp RapidJSON.pc /usr/local/lib/pkgconfig/ \
   && cd / \
   && rm -rf v1.1.0.tar.gz \
--- a/docker/dev/ubi/build.sh
+++ b/docker/dev/ubi/build.sh
@@ -2,11 +2,15 @@
 
 set -x
 
-docker build --build-arg HTTP_PROXY=$http_proxy --build-arg HTTPS_PROXY=$https_proxy --build-arg NO_PROXY=$no_proxy -t noirolabs/ubibase-opflex-build-base -f Dockerfile-ubibase-opflex-build-base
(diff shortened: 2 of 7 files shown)


docker/Dockerfile-opflex-build-base
```diff
@@ -5,13 +5,30 @@ RUN microdnf install --enablerepo codeready-builder-for-rhel-8-x86_64-rpms \
     libtool pkgconfig autoconf automake make cmake file python2-six \
     openssl-devel git gcc gcc-c++ boost-devel diffutils python2-devel \
     libnetfilter_conntrack-devel wget which curl-devel procps zlib-devel \
+    libmnl-devel vi \
     && microdnf clean all
+RUN git clone https://git.netfilter.org/libnftnl \
```
tomflynn, line 10: Can we pull specific tags for libnftlnl and iptables instead of tracking master?

docker/Dockerfile-opflex-build-base
```diff
@@ -5,13 +5,30 @@ RUN microdnf install --enablerepo codeready-builder-for-rhel-8-x86_64-rpms \
...
+  && rm -Rf libnftnl \
+  && git clone https://git.netfilter.org/iptables \
+  && cd iptables \
+  && ./autogen.sh \
+  && ./configure --prefix=/usr --sbindir=/sbin \
+  && make $make_args \
+  && make install && make clean \
+  && cd / \
+  && rm -Rf iptables
 RUN wget https://github.com/Tencent/rapidjson/archive/v1.1.0.tar.gz \
```
tomflynn, line 26: can you merge the two RUN commands using &&. That should reduce the size slightly

docker/dev/ubi/build.sh
```diff
@@ -2,11 +2,15 @@
 
 set -x
 
-docker build --build-arg HTTP_PROXY=$http_proxy --build-arg HTTPS_PROXY=$https_proxy --build-arg NO_PROXY=$no_proxy -t noirolabs/ubibase-opflex-build-base -f Dockerfile-ubibase-opflex-build-base
-docker build --build-arg HTTP_PROXY=$http_proxy --build-arg HTTPS_PROXY=$https_proxy --build-arg NO_PROXY=$no_proxy -t noirolabs/ubibase-opflex -f Dockerfile-ubibase-opflex
-docker build --build-arg HTTP_PROXY=$http_proxy --build-arg HTTPS_PROXY=$https_proxy --build-arg NO_PROXY=$no_proxy -t noirolabs/ubibase-aci -f Dockerfile-ubibase-aci
+export http_proxy=http://proxy.esl.cisco.com:80
```
tomflynn, line 5: I don't think we should explicit refer to internal proxies. It should come from env vars

docker/Dockerfile-opflex-build-base
mchalla: Will do. thanks.

docker/dev/ubi/build.sh
mchalla: sure, this script is not used by the code but by me to build base images so we can run the UBI builds outside of the build servers without the certificates. I did not want to set the env variables each time i ran the script. so it was a convenience thing.

docker/Dockerfile-opflex-build-base
mchalla: done and amended the commit.

docker/Dockerfile-opflex-build-base
mchalla: ok. let me check the latest tags and resend.

snaiksat:
where are we getting the iptables-wrapper script from?

mchalla:
> where are we getting the iptables-wrapper script from?

From here. https://github.com/kubernetes-sigs/iptables-wrappers/blob/master/iptables-wrapper-installer.sh

It has a bug though in this section.
            --install /usr/sbin/iptables iptables /usr/sbin/iptables-wrapper 100 \
            --slave /usr/sbin/iptables-restore iptables-restore /usr/sbin/iptables-wrapper \
            --slave /usr/sbin/iptables-save iptables-save /usr/sbin/iptables-wrapper \
            --slave /usr/sbin/ip6tables iptables /usr/sbin/iptables-wrapper \
            --slave /usr/sbin/ip6tables-restore iptables-restore /usr/sbin/iptables-wrapper \
            --slave /usr/sbin/ip6tables-save iptables-save /usr/sbin/iptables-wrapper

our copy does not have that bug.

docker/Dockerfile-opflex-build-base
mchalla: picked libnftnl-1.1.7 and iptables 1.8.5 thanks.

jojimt (approved):
LGTM

mchalla:
Tom / Sumit, if you are ok with the changes I will merge it. This and the support one need to be merged together. thanks.

tomflynn (approved):

PR #659, opened 2020-10-31
supporting the use case where l3out can be part of

jayaramsatya:
diffrent VRF as long as contract scope set accordingly,
and also verifying for only matching POD and Node BD's

### DIFF
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -503,12 +503,10 @@ func (cont *AciController) Run(stopCh <-chan struct{}) {
 
 	cont.log.Debug("UseAPICInstTag set to:", cont.apicConn.UseAPICInstTag)
 	cont.log.Debug("SnatPbrFltrChain set to:", cont.apicConn.SnatPbrFltrChain)
-
-	// Make sure Pod/NodeBDs and AciL3Out are assoicated to same VRF.
+	// Make sure Pod/NodeBDs are assoicated to same VRF.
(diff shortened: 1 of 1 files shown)


pkg/controller/controller.go
```diff
@@ -438,6 +438,12 @@ func (cont *AciController) Run(stopCh <-chan struct{}) {
 		panic(err)
 	}
 
+	// Build Pod/NodeBD Dn's if not defined.
+	if cont.config.AciPodBdDn == "" || cont.config.AciNodeBdDn == "" {
+		cont.config.AciPodBdDn = "uni/tn-" + cont.config.AciPolicyTenant + "/BD-" + cont.config.AciPolicyTenant + "-pod-bd"
+		cont.config.AciNodeBdDn = "uni/tn-" + cont.config.AciPolicyTenant + "/BD-" + cont.config.AciPolicyTenant + "-node-bd"
+	}
```
snaiksat, line 445: Is there a reason we changed this from the way it was before? I believe we were setting to "" in the overlay case and hence wanted to skip any further validation in that case.

jayaramsatya: Ok good point, haven't thought of Overlay case in that case we need to remove this and also introduce the following checks below cont.config.AciPodBdDn != "". and i hope ACC provision will always provides that config value in case of non overlay mode

snaiksat (approved):

PR #661, opened 2020-11-02
Add hooks for adding a mo via crd

jojimt:
Define an interface and access functions to add delete a GBP mo, based on a CRD.

### DIFF
--- a/pkg/gbpserver/crd_mo.go
+++ b/pkg/gbpserver/crd_mo.go
@@ -20,9 +20,9 @@ import (
 	log "github.com/sirupsen/logrus"
 )
 
-// CRDMo acts as the interface between a CRD and the corresponding MO.
+// GBPCustomMo acts as the interface between a CRD and the corresponding MO.
 // Implement this interface in the CRD watcher
-type CRDMo interface {
+type GBPCustomMo interface {
 	Subject() string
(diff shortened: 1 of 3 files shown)


jayaramsatya (approved):
LGTM, Separate commit is required for dynamic detection CRD's similar to ACI controller in GBP server

pkg/gbpserver/crd_mo.go
```diff
@@ -0,0 +1,105 @@
...
+
+package gbpserver
+
+import (
+	log "github.com/sirupsen/logrus"
+)
+
+// CRDMo acts as the interface between a CRD and the corresponding MO.
+// Implement this interface in the CRD watcher
+type CRDMo interface {
```
tomflynn, line 25: Can we rename to something like OpflexMo or GenieMo?

jojimt: That's too generic. All MOs are Opflex/Genie MOs. How about CustomMo or FeatureMo? @tomflynn

tomflynn: I figured this interface *could* be used for any opflex MO but I guess something like CustomOpflexMo is fine.

tomflynn (approved):

PR #666, opened 2020-11-05
Adding the environment variable to detect multus presence in launch-hostagent.sh script

abhis2112:
Go fmt cmd/acicontainersoperator/main.go

### DIFF
--- a/cmd/acicontainersoperator/main.go
+++ b/cmd/acicontainersoperator/main.go
@@ -15,9 +15,9 @@
 package main
 
 import (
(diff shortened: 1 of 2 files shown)


cmd/acicontainersoperator/main.go
```diff
@@ -15,9 +15,9 @@
 package main
 
 import (
-	log "github.com/sirupsen/logrus"
```
jayaramsatya, line 18: nit: can you revert this file

jayaramsatya (approved):
LGTM

PR #669, opened 2020-11-10
Use the passed timeout instead of ignoring

tomflynn:
All callers pass the same timeout anyway
so this doesn't change the existing behavior

### DIFF
--- a/cmd/opflexagentcni/main.go
+++ b/cmd/opflexagentcni/main.go
@@ -115,7 +115,7 @@ func waitForAllNetwork(result *current.Result, id string,
 		if err != nil {
 			log.Error("Could not open netns: ", err)
 		} else {
-			waitForNetwork(netns, result, id, index, 10*time.Second)
+			waitForNetwork(netns, result, id, index, timeout)
 			netns.Close()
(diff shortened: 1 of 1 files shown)


cmd/opflexagentcni/main.go
```diff
@@ -115,7 +115,7 @@ func waitForAllNetwork(result *current.Result, id string,
 		if err != nil {
 			log.Error("Could not open netns: ", err)
 		} else {
-			waitForNetwork(netns, result, id, index, 10*time.Second)
+			waitForNetwork(netns, result, id, index, timeout)
```
snaiksat, line 118: I believe the intention earlier was to wait for at least 10 s.

tomflynn: Sure both existing callers of this function pass 10 sec as the timeout. The behavior is unchanged by this change

tomflynn: I guess an alternative would be to leave hardcoded to 10 sec here and remove the timeout param

snaiksat (approved):

PR #679, opened 2020-11-20
DropLog: log pod-names instead of addresses

shastrinator:
If available, use pod-names instead of addresses
in pod events

### DIFF
--- a/pkg/hostagent/event_poster.go
+++ b/pkg/hostagent/event_poster.go
@@ -109,7 +109,7 @@ func (agent *HostAgent) processPacketEvent(packetEvent PacketEvent, currTime tim
 	if agent.shouldIgnore(packetEvent, currTime) {
 		return nil
 	}
-	var srcEventPosted bool = false
+	var eventDestination *v1.Pod
 	agent.indexMutex.Lock()
 	srcPodKey, srcOk := agent.podIpToName[packetEvent.SourceIP]
 	agent.indexMutex.Unlock()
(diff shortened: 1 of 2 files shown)


pkg/hostagent/event_poster.go
```diff
@@ -173,8 +160,24 @@ func (agent *HostAgent) processPacketEvent(packetEvent PacketEvent, currTime tim
 	if !srcOk && !dstOk {
 		return nil
 	}
-
-	// update poster's eventSubmitTimeMap
-	agent.poster.eventSubmitTimeMap[packetEvent.SourceIP+packetEvent.DestinationIP] = currTime
+	if eventDestination != nil {
+		var eventSrcName, eventDstName string = packetEvent.SourceIP, packetEvent.DestinationIP
```
jayaramsatya, line 164: Nit: Above lines 160-162, you are checking for srcOk and dskOk  if they are not present you are returning from there,    then you can  directly set the eventSrc and dst names right ?

pkg/hostagent/event_poster.go
```diff
@@ -109,7 +109,7 @@ func (agent *HostAgent) processPacketEvent(packetEvent PacketEvent, currTime tim
 	if agent.shouldIgnore(packetEvent, currTime) {
 		return nil
 	}
-	var srcEventPosted bool = false
+	var eventDestination *v1.Pod = nil
```
jayaramsatya, line 112: nit: you can assume it is nil unless you set the pointer

pkg/hostagent/event_poster.go
shastrinator: The pods may be present(srcOk,dstOk), but not in running state, we don't post events in that case.

pkg/hostagent/event_poster.go
shastrinator: ok, will remove the init.

jayaramsatya (approved):
LGTM

shastrinator:
I merged this a little earlier than the checks passed, sorry about that. I noticed this later.

PR #681, opened 2020-11-23
Minor log enhancements

tomflynn:
Add additional detail to logs that would have helped recent investigation
Use passed timeout in test code


jayaramsatya (approved):
LGTM

PR #686, opened 2020-11-25
Netflow CRD: handle empty attributes checks

tanyatukade:
- Added checks to handle empty attributes. If empty optional attribute detected, it sets to default values defined by apic docs.
- Added few more test cases in test file.
- Changed attribute name "type" (version) to "flowType" to avoid confusion with yaml native type.

### DIFF
--- a/pkg/controller/netflow.go
+++ b/pkg/controller/netflow.go
@@ -184,7 +184,7 @@ func (cont *AciController) handleNetflowPolUpdate(obj interface{}) bool {
 	}
 	apicSlice = append(apicSlice, VmmVSwitch)
 
-	cont.log.Info("create netflow Rs", apicSlice)
+	cont.log.Info("Netflow Object", apicSlice)
 
 	cont.apicConn.WriteApicObjects(labelKey, apicSlice)
 
--- a/pkg/controller/netflow.go
+++ b/pkg/controller/netflow.go
@@ -150,20 +150,38 @@ func (cont *AciController) handleNetflowPolUpdate(obj interface{}) bool {
 	nfDn := nf.GetDn()
 	apicSlice := apicapi.ApicSlice{nf}
 	nf.SetAttr("dstAddr", nfp.Spec.FlowSamplingPolicy.DstAddr)
-	nf.SetAttr("dstPort", strconv.Itoa(nfp.Spec.FlowSamplingPolicy.DstPort))
+
+	if nfp.Spec.FlowSamplingPolicy.DstPort != 0 {
+		nf.SetAttr("dstPort", strconv.Itoa(nfp.Spec.FlowSamplingPolicy.DstPort))
+	} else {
+		nf.SetAttr("dstPort", "unspecified")
+	}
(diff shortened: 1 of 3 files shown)


pkg/controller/netflow.go
```diff
@@ -150,23 +150,41 @@ func (cont *AciController) handleNetflowPolUpdate(obj interface{}) bool {
...
 	VmmVSwitch := apicapi.NewVmmVSwitchPolicyCont(cont.vmmDomainProvider(), cont.config.AciVmmDomain)
 	RsVmmVSwitch := apicapi.NewVmmRsVswitchExporterPol(cont.vmmDomainProvider(), cont.config.AciVmmDomain, nfDn)
 	VmmVSwitch.AddChild(RsVmmVSwitch)
-	RsVmmVSwitch.SetAttr("activeFlowTimeOut", strconv.Itoa(nfp.Spec.FlowSamplingPolicy.ActiveFlowTimeOut))
-	RsVmmVSwitch.SetAttr("idleFlowTimeOut", strconv.Itoa(nfp.Spec.FlowSamplingPolicy.IdleFlowTimeOut))
-	RsVmmVSwitch.SetAttr("samplingRate", strconv.Itoa(nfp.Spec.FlowSamplingPolicy.SamplingRate))
+	if nfp.Spec.FlowSamplingPolicy.ActiveFlowTimeOut != 0 {
+		RsVmmVSwitch.SetAttr("activeFlowTimeOut", strconv.Itoa(nfp.Spec.FlowSamplingPolicy.ActiveFlowTimeOut))
+	} else {
+		RsVmmVSwitch.SetAttr("activeFlowTimeOut", "60")
```
tomflynn, line 173: Do you need the else for these properties? The default is already set to this value anyway in apic_metadata.go

pkg/controller/netflow_test.go
```diff
@@ -124,24 +124,40 @@ func (cont *AciController) netflowPolUpdate(nfp *netflowpolicy.NetflowPolicy) ap
 	nf.SetAttr("dstAddr", nfp.Spec.FlowSamplingPolicy.DstAddr)
 	if nfp.Spec.FlowSamplingPolicy.Version == "netflow" {
 		nf.SetAttr("ver", "v5")
-	}
-	if nfp.Spec.FlowSamplingPolicy.Version == "ipfix" {
+	} else if nfp.Spec.FlowSamplingPolicy.Version == "ipfix" {
```
tomflynn, line 127: I notice that this func is duplicating most of what handleNetflowPolUpdate does in netflow.go. Can you rework the code to exercise handleNetflowPolUpdate instead?

tomflynn (changes requested):

pkg/controller/netflow.go
tanyatukade: The apic_metadata.go has the info for default values but the setAttr is the function that actually sets the attribute with the value.

pkg/controller/netflow_test.go
tanyatukade: Will work on that.

tomflynn (approved):

PR #687, opened 2020-11-25
Serialize podif deletes with adds

jojimt:
This is to make sure that deletes are not attempted before a create completes, leaving stale entries as seen in a kind cluster.


tomflynn (approved):

jayaramsatya (approved):
LGTM

PR #704, opened 2020-12-14
Add UTs for Netflow CRD

tanyatukade:
- Added UTs for Netflow CRD to increase coverage

### DIFF
--- a/pkg/controller/netflow.go
+++ b/pkg/controller/netflow.go
@@ -80,15 +80,16 @@ func (cont *AciController) initNetflowInformerBase(listWatch *cache.ListWatch) {
 
 }
 
-func (cont *AciController) netflowPolicyUpdated(obj interface{}) {
+func (cont *AciController) netflowPolicyUpdated(obj interface{}) bool {
 	netflowPolicy := obj.(*netflowpolicy.NetflowPolicy)
 	key, err := cache.MetaNamespaceKeyFunc(netflowPolicy)
 	if err != nil {
 		NetflowPolicyLogger(cont.log, netflowPolicy).
 			Error("Could not create key:" + err.Error())
-		return
+		return false
 	}
 	cont.queueNetflowUpdateByKey(key)
+	return false
 
 }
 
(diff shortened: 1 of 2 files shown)


pkg/controller/netflow.go
```diff
@@ -80,15 +80,16 @@ func (cont *AciController) initNetflowInformerBase(listWatch *cache.ListWatch) {
...
 	netflowPolicy := obj.(*netflowpolicy.NetflowPolicy)
 	key, err := cache.MetaNamespaceKeyFunc(netflowPolicy)
 	if err != nil {
 		NetflowPolicyLogger(cont.log, netflowPolicy).
 			Error("Could not create key:" + err.Error())
-		return
+		return false
 	}
 	cont.queueNetflowUpdateByKey(key)
+	return false
```
tomflynn, line 92: Shouldn't this return true? Right now it always returns false...

pkg/controller/netflow.go
```diff
@@ -106,44 +107,38 @@ func (cont *AciController) queueNetflowUpdate(netflowpolicy *netflowpolicy.Netfl
...
 	}
 	nfkey, err := cache.MetaNamespaceKeyFunc(nf)
 	if err != nil {
 		NetflowPolicyLogger(cont.log, nf).
 			Error("Could not create netflow key: ", err)
-		return
+		return false
 	}
 	cont.apicConn.ClearApicObjects(cont.aciNameForKey("nfp", nfkey))
+	return false
```
tomflynn, line 133: same comment...you're always returning false regardless of whether there's a problem

tomflynn (changes requested):

pkg/controller/netflow.go
tanyatukade: You are right! I have addressed your comments in the commit below.

jayaramsatya: you don't need a return bool value here.  make it void function you are not handling any error condition here.

tanyatukade: changed it.

tomflynn (approved):

PR #706, opened 2020-12-15
Fix requeuing of netflow

tanyatukade:


pkg/controller/netflow.go
```diff
@@ -199,5 +199,5 @@ func (cont *AciController) handleNetflowPolUpdate(obj interface{}) bool {
 	labelKey := cont.aciNameForKey("nfp", key)
 	cont.apicConn.WriteApicObjects(labelKey, cont.netflowPolObjs(nfp))
 
-	return true
+	return false
```
tomflynn, line 202: Can you add a comment to the top of the function indicating what the return value indicates? (true - caller should retry/requeue, false - done)

tomflynn (changes requested):

pkg/controller/netflow.go
tanyatukade: added

tomflynn (approved):

PR #712, opened 2020-12-17
Add Erspan CRD for non-overlay case

tanyatukade:
This patch adds the ability to create ERSPAN sessions on k8s ACI CNI for non-overlay case.

- Added Podif CRD in controller. The information from podif is used to configure ERSPAN source. 
- Tested on fabric for create/update/delete.
- Policy config is received by OVS.
- Tested POST to APIC.

Signed-off-by: Tanya Tukade <tanyatukade.123@gmail.com>


pkg/hostagent/pods.go
```diff
@@ -355,7 +355,7 @@ func (agent *HostAgent) syncEps() bool {
 				if err != nil {
 					opflexEpLogger(agent.log, ep).
 						Error("Error writing EP file: ", err)
-				} else if wrote || !ep.registered {
+				} else if wrote {
```
tomflynn, line 358: Why was this check removed as part of this change?

pkg/controller/services.go
```diff
@@ -984,6 +984,8 @@ func (cont *AciController) opflexDeviceChanged(obj apicapi.ApicObject) {
 
 		for _, node := range nodeUpdates {
 			cont.env.NodeServiceChanged(node)
+			cont.erspanFullSync()
```
tomflynn, line 987: Why would this be called multiple times in a loop?

tomflynn (changes requested):
I haven't looked closely at this PR yet but we need to enhance the UTs. code coverage in erspan.go is < 30%

pkg/hostagent/pods.go
tanyatukade: This was initially removed for testing creation of podif CRD in controller,  since ep-registry in nil for on prem case. @jayaramsatya Can you please comment on this? Now that we have the flag for on prem case turned on in ep-registry, should I revert this check?

jayaramsatya: initially just to test the pod if creates I have asked to remove this. now that you have enabled the podif creations default you can revert this change

tanyatukade: Thanks @jayaramsatya

pkg/controller/services.go
tanyatukade: You are right, this is not needed. I have called the erspanFullSync in environment.go

pkg/controller/erspan.go
```diff
@@ -0,0 +1,386 @@
...
+	cont.log.Debug("erspan key queued: ", key)
+	cont.erspanQueue.Add(key)
+}
+
+func (cont *AciController) queuePodIfUpdateByKey(key string) {
+	cont.log.Debug("Podif key queued: ", key)
+	cont.podIfQueue.Add(key)
+}
+
+func (cont *AciController) handlePodifInfo(podif *podIfpolicy.PodIF) bool {
```
tomflynn, line 202: Do we really need this? It looks like this does a lookup and logs differently based on whether podif is already in the map or not...that doesn't seem worth the handling. Can we remove this or am I missing something?

tomflynn (changes requested):
Still reviewing but wanted to provide some feedback

pkg/controller/controller.go
```diff
@@ -187,6 +196,13 @@ type portRangeSnat struct {
 	end   int
 }
 
+//EndPointData holds PodIF data in controller.
+type EndPointData struct {
+	MacAddr   string
+	EPG       string
```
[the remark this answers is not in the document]
tanyatukade: I am splitting the EPG in erspan.go when I locally cache it. It is inside podIFAdded func.

pkg/controller/erspan.go
tanyatukade: We need to add podif obj to the queue in the environment.go and process it. Otherwise the podif crd wont be processed in the controller. We can remove the remove the if else statements and have an empty func.

pkg/controller/config.go
```diff
@@ -107,6 +107,9 @@ type ControllerConfig struct {
 	// Tenant to use when creating policy objects in APIC
 	AciPolicyTenant string `json:"aci-policy-tenant,omitempty"`
 
+	// App-Profile to use when creating fvCEp objects in APIC
+	AciAppProfile string `json:"aci-app-profile,omitempty"`
```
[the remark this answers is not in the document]
tanyatukade: shouldnt the app profile be injected in the deployment file as an input?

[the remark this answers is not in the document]
tomflynn: The app profile could be but my point is that the user is allowed to annotate pods and move them into other app-profile/epgs. This information is available in podif so always getting the app profile from there is safest and will handle both cases

pkg/controller/controller.go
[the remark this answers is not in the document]
tomflynn: Yes but you're dropping the app-profile aren't you?

pkg/controller/config.go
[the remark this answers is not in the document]
tanyatukade: Got it, will handle that

pkg/controller/controller.go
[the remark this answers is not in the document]
tanyatukade: Yes, will make sure app profile is cached as well to use for fvCEp dn.

pkg/controller/erspan.go
tomflynn: Are you sure about that? I don't see why you need the queue at all. Your callback isn't doing anything and I don't see any side effects in the processQueue handling that would be needed. The calls to queuePodIfUpdateByKey also don't seem to be needed

jayaramsatya: we don't require a queue here this can be removed. you can handle it directly Podif update function

pkg/controller/pods.go
```diff
@@ -135,6 +135,7 @@ func (cont *AciController) podAdded(obj interface{}) {
 	pod := obj.(*v1.Pod)
 	cont.writeApicPod(pod)
 	cont.depPods.UpdatePodNoCallback(pod)
+	cont.erspanPolPods.UpdatePodNoCallback(pod)
```
tomflynn, line 138: Why do you need to react to pod events? I thought ERSPAN would only react to podif events

tanyatukade: Yes, you are right. It reacts to podif events. Will clean it up.

pkg/controller/environment.go
```diff
@@ -270,6 +277,17 @@ func (env *K8sEnvironment) PrepareRun(stopCh <-chan struct{}) error {
 		func(obj interface{}) bool {
 			return cont.handleSnatNodeInfo(obj.(*snatnodeinfo.NodeInfo))
 		}, stopCh)
+	go cont.processQueue(cont.erspanQueue, cont.erspanIndexer,
+		func(obj interface{}) bool {
+			return cont.handleErspanUpdate(obj.(*erspanpolicy.ErspanPolicy))
+		}, stopCh)
+	cont.log.Debug("Waiting for erspan cache sync")
+	cont.erspanFullSync()
```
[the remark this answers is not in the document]
tanyatukade: Sure, It can be triggered in erspanInit. Addressed it.

pkg/controller/pods.go
jayaramsatya: Tom, here this change is required to select the pods based on the labels  coming from ERSPAN CR. after selecting the pods,  using the pod info  podif cache can be used to get the required info

pkg/controller/environment.go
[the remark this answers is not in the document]
tanyatukade: Since we are not doing a GET from APIC for the erspan sync, I am changing the func name.

pkg/controller/erspan.go
```diff
@@ -0,0 +1,358 @@
...
+func getPodifKey(podif *podIfpolicy.PodIF) string {
+	return podif.Status.PodNS + "/" + podif.Status.PodName
+}
+
+func (cont *AciController) podIFAdded(obj interface{}) {
+	podif := obj.(*podIfpolicy.PodIF)
+	cont.log.Infof("podif Added: %s", podif.ObjectMeta.Name)
+	cont.indexMutex.Lock()
+	deconcatenate := strings.Split(podif.Status.EPG, "|")
+	deconEPG := deconcatenate[len(deconcatenate)-1]
```
tomflynn, line 144: Although it shouldn't happen, we should add some error checking here in case the EPG name isn't of the expected form to avoid a runtime error

pkg/controller/erspan.go
```diff
@@ -0,0 +1,358 @@
...
+func (cont *AciController) handleErspanUpdate(span *erspanpolicy.ErspanPolicy) bool {
+
+	erspanLogger := ErspanPolicyLogger(cont.log, span)
+	spankey, err := cache.MetaNamespaceKeyFunc(span)
+	if err != nil {
+		erspanLogger.Error("Could not create erspan policy key: ", err)
+		return false
+	}
+	labelKey := cont.aciNameForKey("span", spankey)
+	cont.log.Debug("creating erspanpolicy")
```
tomflynn, line 266: can you add the policy name to the log as well?

pkg/controller/erspan.go
```diff
@@ -0,0 +1,358 @@
...
+	if span.Spec.Source.Direction != "" {
+		src.SetAttr("dir", span.Spec.Source.Direction)
+	} else {
+		src.SetAttr("dir", "both")
+	}
+
+	// Build fvCEp for matching pods
+	cont.indexMutex.Lock()
+	podKeys := cont.erspanPolPods.GetPodForObj(spankey)
+	cont.log.Debug("podkeys are: ", podKeys)
```
tomflynn, line 288: This could be pretty verbose. Can you remove? You're logging the individual items in the loop as well

pkg/controller/erspan.go
```diff
@@ -0,0 +1,358 @@
...
+	for node := range cont.nodeOpflexDevice {
+		fabricPath, ok := cont.fabricPathForNode(node)
+		if !ok {
+			continue
+		}
+		paths = append(paths, fabricPath)
+	}
+	cont.indexMutex.Unlock()
+
+	var vpcs []string
```
tomflynn, line 332: What if the opflexODev is on an access port?

tanyatukade: Added support for Leaf Access Ports in the latest commit

pkg/controller/erspan.go
```diff
@@ -0,0 +1,428 @@
...
+		cont.log.Infof("Invalid EPG: %s", podif.Status.EPG)
+	}
+	cont.indexMutex.Unlock()
+}
+
+func (cont *AciController) podIFDeleted(obj interface{}) {
+	podif := obj.(*podIfpolicy.PodIF)
+	cont.log.Infof("podif Deleted: %s", podif.ObjectMeta.Name)
+	cont.indexMutex.Lock()
+	delete(cont.podIftoEp, podif.ObjectMeta.Name)
```
jayaramsatya, line 175: This will leak here Key should be getPodifKey(podif)

pkg/controller/erspan.go
```diff
@@ -0,0 +1,428 @@
...
+	if err != nil {
+		ErspanPolicyLogger(cont.log, erspanPolicy).
+			Error("Could not create key:" + err.Error())
+		return
+	}
+	cont.erspanPolPods.UpdateSelectorObj(obj)
+	cont.queueErspanUpdateByKey(spankey)
+}
+
+func (cont *AciController) queueErspanUpdateByKey(key string) {
```
jayaramsatya, line 191: Nit: we might not require a queue here as we are not reconciling on failure

pkg/controller/erspan.go
```diff
@@ -0,0 +1,428 @@
...
+func isValidEPG(podif *podIfpolicy.PodIF) bool {
+	pattern := regexp.MustCompile("\\|")
+	idx := pattern.FindAllStringIndex(podif.Status.EPG, -1)
+	if len(idx) != 1 || idx[0][0] == 0 || idx[0][0] == len(podif.Status.EPG)-1 {
+		return false
+	}
+	return true
+}
+
+func (cont *AciController) podIFAdded(obj interface{}) {
```
jayaramsatya, line 149: How are we reconciling the ERSPAN configuration if there is any PODIF data changes?

pkg/controller/erspan_test.go
```diff
@@ -0,0 +1,321 @@
...
+				spanTest.erspanPol.Namespace+"_"+spanTest.erspanPol.Name)
+			if !tu.WaitEqual(t, last, 0,
+				len(cont.apicConn.GetDesiredState(key)), "delete") {
+				return false, nil
+			}
+			return true, nil
+		})
+}
+
+func addpods(cont *testAciController, incIps bool, ips []string, ipv4 bool) {
```
jayaramsatya, line 144: are you not able to refer it from the Network Policy test code, if required move it to common.

pkg/controller/erspan_test.go
```diff
@@ -0,0 +1,321 @@
...
+	// podifwait(t, "podif data", expected, cont.AciController.podIftoEp)
+	// podifwait(t, "podif data", expected, cont.AciController.podIftoEp)
+	cont.stop()
+
+	for _, spanTest := range spanTests {
+		cont := initCont()
+		cont.log.Info("Testing erspan post to APIC ", spanTest.desc)
+		addpods(cont, true, ips, true)
+		cont.run()
+		cont.fakeNetflowPolicySource.Modify(spanTest.erspanPol)
```
jayaramsatya, line 308: why you are referring to NetflowPolicy here

pkg/controller/erspan_test.go
```diff
@@ -0,0 +1,321 @@
...
+		erspanObject(t, "object absent check", cont, name, "spanVSrcGrp", false)
+		erspanObject(t, "object absent check", cont, name, "spanVDestGrp", false)
+		actualPost := spanTest.writeToApic
+		expectedPost := cont.handleErspanUpdate(spanTest.erspanPol)
```
jayaramsatya, line 312: don't call the function directly, this has to be triggered by watchers

pkg/controller/erspan.go
```diff
@@ -0,0 +1,428 @@
...
+	podif := obj.(*podIfpolicy.PodIF)
+	cont.log.Infof("podif Added: %s", podif.ObjectMeta.Name)
+	cont.indexMutex.Lock()
+	if isValidEPG(podif) {
```
jayaramsatya, line 153: can we fix this issue if there is any EPG issue before populating the podif's

pkg/controller/erspan.go
tanyatukade: Good catch, will fix it

pkg/controller/erspan.go
tanyatukade: That makes sense, will remove the check here.

pkg/controller/erspan.go
```diff
@@ -0,0 +1,428 @@
...
+	cont.erspanPolPods = index.NewPodSelectorIndex(cont.log,
+		cont.podIndexer, cont.namespaceIndexer, cont.erspanIndexer,
+		cache.MetaNamespaceKeyFunc,
+		func(obj interface{}) []index.PodSelector {
+			span := obj.(*erspanpolicy.ErspanPolicy)
+			ls := &metav1.LabelSelector{MatchLabels: span.Spec.Selector.Labels}
+			return index.PodSelectorFromNsAndSelector(span.Spec.Selector.Namespace, ls)
+		},
+	)
+	cont.erspanPolPods.SetPodUpdateCallback(func(podkey string) {
```
jayaramsatya, line 251: do we need this change?  calling pod update  we are not doing anything specific to ERSPAN, may 
be we can remove this code

tanyatukade: Yes, you are right, will clean this up, thanks

pkg/controller/erspan.go
tanyatukade: Will work on this as a separate PR.

jayaramsatya (approved):
As we discussed Please Add the Support issue for PODIF changes and also add more UT's.
Other Changes LGTM

tomflynn (approved):

PR #715, opened 2021-01-12
Uuid package is changed to github.com/google/uuid

jayaramsatya:


tomflynn (dismissed):
Looks fine but can we also remove vendor/github.com/nu7hatch and update go.mod + vendor/modules.txt

tomflynn (approved):

PR #717, opened 2021-01-14
ACI CNI base images

snaiksat:
We are introducing the following layered hierarchy for
the ACI CNI images:

aci-containers-base
aci-containers-\<component\>-base
aci-containers-\<component\>

with the goal of:

a. Normalizing the ACI CNI common packages
of the current images into a common base, and
such additional packages can be added in one
common image layer as opposed to being added
independently in each component image (this goes
into the aci-containers-base image)

b. Identifying the static parts of the component-
specific images such that they are not required to
pulled everytime (this goes into the
aci-containers-\<component\>-base image which builds
on top of the aci-containers-base image)

The ACI CNI compiled artifacts go into the final
aci-containers-\<component\> image which builds
on top of the above two.

The base images need to be built less frequently
and can be cached thus substantially reducing the
final image build time in the CI setup.

This approach is similar to what has already been
adopted for the opflex container.

In this PR only the Dockerfiles for images used in the
on-prem deployment are being updated.

### DIFF
--- a/docker/Dockerfile-aci-containers-base
+++ b/docker/Dockerfile-aci-containers-base
@@ -1,0 +1,3 @@
+FROM registry.access.redhat.com/ubi8/ubi:latest
+RUN yum --disablerepo=\*ubi\* install -y curl
+CMD ["/usr/bin/sh"]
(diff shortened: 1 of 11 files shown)


jojimt:
We also need to cover the gbp-server images here, right?

snaiksat:
@jojimt , yes, I will do in a follow-up PR.

jayaramsatya (approved):
LGTM

PR #745, opened 2021-03-03
controller is crashing due to epspan sync function is called before the erspan informers are in sync

jayaramsatya:
Changed the sequence of crds calling

Revert to goroutine when CRDs are added

Run goroutine after crd registration

### DIFF
--- a/pkg/controller/environment.go
+++ b/pkg/controller/environment.go
@@ -18,7 +18,6 @@ import (
 	"context"
 	"strings"
 
-	erspanpolicy "github.com/noironetworks/aci-containers/pkg/erspanpolicy/apis/aci.erspan/v1alpha"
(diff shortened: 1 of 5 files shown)


pkg/controller/environment.go
```diff
@@ -265,20 +258,11 @@ func (env *K8sEnvironment) PrepareRun(stopCh <-chan struct{}) error {
 		func(obj interface{}) bool {
 			return cont.handleQosPolUpdate(obj)
 		}, stopCh)
```
tanyatukade, line 260: Shouldnt the Qos queue be processed after the qosInformer has run in qos.go?

jayaramsatya: Yes, we need to move this

jayaramsatya: I have moved the function.

tanyatukade (approved):

PR #746, opened 2021-03-04
ERSPAN sync with opflex devices

tanyatukade:
- ERSPAN object needs to sync with every opflex device change, since it uses the vpcs to build Rs objects.
- Added error checking when erspanIndexer is nil.


jayaramsatya (approved):
LGTM

tomflynn (approved):

PR #752, opened 2021-03-14
Assign consistent mac address based on the pod's Ip address

Bhavanaashok33:
Related GitHub issue: https://github.com/noironetworks/support/issues/1391  
Signed-off-by: bashokba <bashokba@cisco.com>

### DIFF
--- a/pkg/hostagent/setup.go
+++ b/pkg/hostagent/setup.go
@@ -17,6 +17,7 @@ package hostagent
 import (
 	"errors"
 	"fmt"
+	"net"
 	"os"
 
 	cnicur "github.com/containernetworking/cni/pkg/types/current"
@@ -80,6 +81,7 @@ type SetupVethArgs struct {
 	Sandbox string
 	IfName  string
 	Mtu     int
+	Ip      net.IP
 }
 
 type SetupVethResult struct {
@@ -88,10 +90,10 @@ type SetupVethResult struct {
 }
 
 func runSetupVeth(sandbox string, ifName string,
-	mtu int) (string, string, error) {
+	mtu int, ip net.IP) (string, string, error) {
 	result := &SetupVethResult{}
 	err := PluginCloner.runPluginCmd("ClientRPC.SetupVeth",
-		&SetupVethArgs{sandbox, ifName, mtu}, result)
+		&SetupVethArgs{sandbox, ifName, mtu, ip}, result)
 	return result.HostVethName, result.Mac, err
 }
 
@@ -111,6 +113,17 @@ func (*ClientRPC) SetupVeth(args *SetupVethArgs, result *SetupVethResult) error 
 		}
 
 		result.HostVethName = hostVeth.Name
+
+		// Force a consistent MAC address based on the IPv4 address
+		// Currently we dont have a support for V6 based mac allocation. Upstream doesn't support yet :-(
+		// This code has to be revisted if upstream drops SetHWAddrByIP in future
+		// https://github.com/containernetworking/plugins/blob/e1517e2498fe4774435bc4be6bdd39fa735b469b/pkg/ip/link_linux.go#L228
+
+		if args.Ip.To4() != nil {
+			if err := ip.SetHWAddrByIP(args.IfName, args.Ip, nil); err != nil {
+				return fmt.Errorf("failed Ip based MAC address allocation for v4: %v", err)
+			}
+		}
 
 		contVeth, err := netlink.LinkByName(args.IfName)
 		if err != nil {
@@ -268,11 +281,6 @@ func (agent *HostAgent) configureContainerIfaces(metadata *md.ContainerMetadata)
 		} else {
 			mtu = agent.config.InterfaceMtu
 		}
-		iface.HostVethName, iface.Mac, err =
-			runSetupVeth(iface.Sandbox, iface.Name, mtu)
-		if err != nil {
-			return nil, err
-		}
 
 		if len(iface.IPs) == 0 {
 			// We're doing ip address management
@@ -281,6 +289,21 @@ func (agent *HostAgent) configureContainerIfaces(metadata *md.ContainerMetadata)
 			err = agent.allocateIps(iface, podKey)
 			if err != nil {
 				return nil, err
+			}
+		}
+
+		for _, ip := range iface.IPs {
+			//There are 4 cases: IPv4-only, IPv6-only, dual stack with either IPv4 or IPv6 as the first address.
+			//We are guaranteed to derive the MAC address from the IPv4 address in the IPv4-only case.
+
+			if ip.Address.IP != nil {
+				iface.HostVethName, iface.Mac, err =
+					runSetupVeth(iface.Sandbox, iface.Name, mtu, ip.Address.IP)
+				if err != nil {
+					return nil, err
+				} else {
+					break
+				}
 			}
 		}
 


pkg/hostagent/setup.go
```diff
@@ -112,6 +114,11 @@ func (*ClientRPC) SetupVeth(args *SetupVethArgs, result *SetupVethResult) error
 
 		result.HostVethName = hostVeth.Name
 
+		// Force a consistent MAC address based on the IP address
+		if err := ip.SetHWAddrByIP(args.IfName, args.PodIp, nil); err != nil {
```
jayaramsatya, line 118: can you check for ipv6? for now, ignore calling if it is of type V6,  as  API is not supporting V6 we can ignore that, next time support comes it is easy to identify, also put the comment there

pkg/hostagent/setup.go
```diff
@@ -284,6 +286,16 @@ func (agent *HostAgent) configureContainerIfaces(metadata *md.ContainerMetadata)
 			}
 		}
 
+		for _, ip := range iface.IPs {
```
tomflynn, line 289: If we had a dual-stack setup, this would create two veths

pkg/hostagent/setup.go
```diff
@@ -284,6 +286,16 @@ func (agent *HostAgent) configureContainerIfaces(metadata *md.ContainerMetadata)
...
+			if ip.PodIp != nil {
```
tomflynn, line 290: Do we really need a separate PodIp field? My reading of the code is that the existing Address.IP field will get set with the allocated address already. Since you do the IP allocate first now, it will only get this far if an IP has been allocated

Bhavanaashok33: Yes, you are right.  we don't need a separate field. I was under as assumption that the existing Address field is of type net.IPNet which is a combination of address and mask. whereas ip.SetHWAddrByIP takes the IP parameter which is of type "net.IP". Since we can directly access Address.IP , we wouldn't need this extra field.

pkg/hostagent/setup.go
Bhavanaashok33: Ok I will handle the ipv6 case.

snaiksat:
Thanks @Bhavanaashok33  for posting the PR. Given that SetHWAddrByIP only supports IPv4, and to avoid unnecessary complexity in the code, I am just wondering if calling:
https://github.com/containernetworking/plugins/blob/e1517e2498fe4774435bc4be6bdd39fa735b469b/pkg/ip/link_linux.go#L228

after:
https://github.com/noironetworks/aci-containers/blob/master/pkg/hostagent/setup.go#L285

only if IPv4 IP has been allocated, is a better idea. Yes, this will result in the MAC address getting allocated once and then getting overwritten, but it might not be that bad.

Secondly, upstream is not using SetHWAddrByIP anymore. It's possible they might drop this function entirely in the future. We should at least have a comment in our code to revisit (as also for IPv6 processing).

snaiksat:
Ah, scratch my first suggestion, you can't set the MAC address after creating the veth, my bad!

Bhavanaashok33:
code handles both ipv6 and ipv4 processing. Since the upstream doesnt support ipv6 based mac allocation yet, I have added comments for furture reference. The codes has been tested on the setup and verified ip based mac allocation. Upon one assigning same ip to the pod, mac address assigned as well be consistent with the IP. This would aviod arp cache mis behavior

pkg/hostagent/setup.go
Bhavanaashok33: for the dual-stack ips, the code will exit the loop if  iface.HostVethName and iface.Mac is already assigned

pkg/hostagent/setup.go
```diff
@@ -284,6 +286,16 @@ func (agent *HostAgent) configureContainerIfaces(metadata *md.ContainerMetadata)
...
+				iface.HostVethName, iface.Mac, err =
+					runSetupVeth(iface.Sandbox, iface.Name, mtu, ip.PodIp)
+				if err != nil {
```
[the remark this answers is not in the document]
Bhavanaashok33: Yes, the code has to handle V6. I have incorporated the changes for ipv6 processing

pkg/hostagent/setup.go
```diff
@@ -284,6 +302,19 @@ func (agent *HostAgent) configureContainerIfaces(metadata *md.ContainerMetadata)
 			}
 		}
 
+		for _, ip := range iface.IPs {
+			if iface.HostVethName != "" && iface.Mac != "" {
```
tomflynn, line 306: why not just have an else case on the err != nil check that breaks out of the loop?

tomflynn (dismissed):
Looks good

jayaramsatya (approved):
LGTM

Bhavanaashok33:
Hi @snaiksat Can you take a look at the  latest changes before it gets merged ?

pkg/hostagent/setup.go
```diff
@@ -284,6 +302,18 @@ func (agent *HostAgent) configureContainerIfaces(metadata *md.ContainerMetadata)
 			}
 		}
 
+		for _, ip := range iface.IPs {
+			if ip.Address.IP != nil {
```
snaiksat, line 306: In the dual stack case, if you see the v6 first, then will we not derive the MAC address based on v4 IP?

pkg/hostagent/setup.go
```diff
@@ -112,6 +114,27 @@ func (*ClientRPC) SetupVeth(args *SetupVethArgs, result *SetupVethResult) error
...
+			{
+				if err := ip.SetHWAddrByIP(args.IfName, args.Ip, nil); err != nil {
+					return fmt.Errorf("failed to set pod interface MAC address for v4: %v", err)
+				}
+			}
+		case args.Ip.To16() != nil:
+			{
+				//TO DO. Upstream doesn't handle ipv6 based mac allocation yet
+			}
+		}
```
snaiksat, line 136: The above switch logic is also present in the SetHWAddrByIP implementation. Perhaps we could just check here if its an ipv4 and then call SetHWAddrByIP? (This is a minor comment, not required to merge this PR, we can do this as a follow-up patch too.)

Bhavanaashok33: That's how the previous implementation looked like where I just checked for ipv4 and called the func SetHWAddrByIP. Since @jayaramsatya had a request to handle ipv6 , I changed the logic to the above one. I can  revert the changes if you insist.

pkg/hostagent/setup.go
Bhavanaashok33: Yes, Thats right. We yet don't support IPv6 isn't it ?

snaiksat: In the dual stack case, the MAC address needs to be derived based on ipv4. Will that happen here if the ipv6 address is processed first (since you break out of the loop once the ipv6 is processed)?

tomflynn: My understanding was that the IPv4 address will always come first in the dual-stack case. That's based on mailing threads I've seen in the past though and I'm not 100% sure it's guaranteed

pkg/hostagent/setup.go
Bhavanaashok33: Just confirmed with Jayaram. The ask was to handle ipv6 by printing a log instead of siliently ignoring.  I will revert the changes and will switch to If condition based logic

pkg/hostagent/setup.go
Bhavanaashok33: To avoid any confusions, I will have a check on ipv4 and on ipv6. If Ipv6, I will continue with the for loop and if ipv4, I will call runSetupVeth.

tomflynn: That would break if we ever need to support PODs with only IPv6 addresses

Bhavanaashok33: How about handling ipv4 by assigning mac address based on IP, and for ipv6 retain the current implementation (Once the upstream supports ipv6 based mac allocation, we can handle it by calling ip based mac alloc)   ?

tomflynn: Sure that's fine

Bhavanaashok33: ok thanks Tom

Bhavanaashok33: so 4 cases: 
1. ipv4 only
2. ipv6 only
3. ipv6 and ipv4 
4. ipv4 and ip6 

step1. if ipv4 only, check for ipv4 -> call runSetupVeth -> assign mac based on ipv4 ->  assign the results i.e result.hostveth and result.mac 
step2. if ipv6 only, check for ipv6 > call runSetupVeth -> “NO” mac based ip assignment > follow the current implementation
step3: for ipv6 and ipv4 , if ipv6 comes first > repeat step2 > for the next iteration, check if veth pair already exists. if yes, break the loop (this is to avoid creating two veths for dual stack setup)
step4: for ipv4 and ipv6, if ipv4 comes first -> repeat step1 >   for the next iteration, check if veth pair already exists. if yes, break the loop

Bhavanaashok33: I will proceed with this logic if no concerns

Bhavanaashok33:
Testing is in process !

Bhavanaashok33:
While working with dual-stack setup on a cluster, seems like the code requires more testing. Acc-provision changes has to go in as well!  The code has been tested with IPv4. Please proceed with the review for now!

pkg/hostagent/setup.go
```diff
@@ -284,6 +292,21 @@ func (agent *HostAgent) configureContainerIfaces(metadata *md.ContainerMetadata)
 			}
 		}
 
+		for _, ip := range iface.IPs {
+			// We are handling 4 cases  here: ipv4 only, ipv6 only, ipv6 and ipv4, and ipv4 and ipv6
```
tomflynn, line 296: I don't think this comment applies with the current code

pkg/hostagent/setup.go
```diff
@@ -284,6 +292,21 @@ func (agent *HostAgent) configureContainerIfaces(metadata *md.ContainerMetadata)
...
+			if ip.Address.IP != nil {
+				if iface.HostVethName != "" && iface.Mac != "" {
```
tomflynn, line 298: Nit: I don't think we really need this check

can't we simplify the block with
```
iface.HostVethName, iface.Mac, err =
    runSetupVeth(iface.Sandbox, iface.Name, mtu, ip.Address.IP)

if err != nil {
    return nil, err
} else {
    break
}
```

pkg/hostagent/setup.go
snaiksat: My suggestion would be to say:
There are 4 cases: IPv4-only, IPv6-only, dual stack with either IPv4 or IPv6 as the first address. We are guaranteed to derive the MAC address from the  IPv4 address in the IPv4-only case.

Bhavanaashok33: Okay sounds Good

pkg/hostagent/setup.go
Bhavanaashok33: Okay sure. With the above changes where we allocate mac address based on ipv4 only , and the rest remain same for Ipv6, this should handle both ipv4 only, ipv6 only, dual-stack with ipv6 first or idual stack Ipv4 first.

tomflynn (approved):

snaiksat (approved):

PR #753, opened 2021-03-15
Allocating multiple IPv6 addrs per pod

tomflynn:
Issue was introduced by an earlier refactoring. IPv4 allocation
was updated correctly but IPv6 allocation uses both new/old styles


jayaramsatya (approved):
LGTM

PR #758, opened 2021-03-27
Added changes to reconcile the topology aware services if the nodes label changes.

jayaramsatya:


pkg/hostagent/nodes.go
```diff
@@ -73,19 +73,28 @@ func (agent *HostAgent) initNodeInformerBase(listWatch *cache.ListWatch) {
...
 	updateServices := false
-
-	node := obj.(*v1.Node)
+	var node *v1.Node
+	if len(obj) == 2 {
+		oldnode := obj[0].(*v1.Node)
+		node = obj[1].(*v1.Node)
+		if !reflect.DeepEqual(node.ObjectMeta.Labels, oldnode.ObjectMeta.Labels) {
+			updateServices = true
+			agent.log.Info("Node label changed Update topology aware services")
```
snaiksat, line 93: By setting this flag we are going to call updateAllServices(). I am guessing that should be good. If so, does this log need to change to just say - "Node label changed, updating services"?

jayaramsatya: i will update the log

snaiksat (approved):

PR #766, opened 2021-04-08
Added support for session Affinity for services.

jayaramsatya:
Session affinity always directs traffic from a client to the same pod.
It is typically used as an optimization to ensure that the same pod receives traffic from the same user so that you can leverage session caching

### DIFF
--- a/pkg/hostagent/services.go
+++ b/pkg/hostagent/services.go
@@ -49,8 +49,9 @@ type opflexServiceMapping struct {
 	NextHopIps  []string `json:"next-hop-ips"`
 	NextHopPort uint16   `json:"next-hop-port,omitempty"`
 
-	Conntrack bool   `json:"conntrack-enabled"`
-	NodePort  uint16 `json:"node-port,omitempty"`
+	Conntrack       bool                         `json:"conntrack-enabled"`
+	NodePort        uint16                       `json:"node-port,omitempty"`
+	SessionAffinity *opflexSessionAffinityConfig `json:"session-affinity,omitempty"`
 }
 
 type opflexService struct {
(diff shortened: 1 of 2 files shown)


pkg/hostagent/services.go
```diff
@@ -616,7 +633,16 @@ func (sep *serviceEndpoint) SetOpflexService(ofas *opflexService, as *v1.Service
 	return hasValidMapping
 }
 
-func CheckKeyMatch(topology, nodelabels map[string]string, key string) bool {
+func getSessionAffinity(config *v1.SessionAffinityConfig) *opflexSessionAffinityConfig {
+	if config.ClientIP != nil && config.ClientIP.TimeoutSeconds != nil {
+		return &opflexSessionAffinityConfig{ClientIP: opflexClientIPConfig{TimeoutSeconds: *config.ClientIP.TimeoutSeconds}}
+	} else {
+		// Default value is 10800(for 3 hours)
+		return &opflexSessionAffinityConfig{ClientIP: opflexClientIPConfig{TimeoutSeconds: 10800}}
```
snaiksat, line 641: Could we set this as a constant such that its easy to update later?

snaiksat (commented):
Thanks @jayaramsatya , could you please update the commit message with a link to the API reference:
https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.19/#service-v1-core
 
and the following example:
kind: Service
apiVersion: v1
metadata:
  name: my-service
spec:
  selector:
    app: my-app
  ports:
  - name: http
    protocol: TCP
    port: 80
    targetPort: 80
  sessionAffinity: ClientIP
  sessionAffinityConfig:
    clientIP:
      timeoutSeconds: 10000

snaiksat (approved):

PR #768, opened 2021-04-09
CSCvx38027 Allow labels to be injected for k8s namespaces/services

mpaidipa-aci:
Changes to push vmmInjectedLabel for Ns, Rs, depl, pod and svc

### DIFF
--- a/pkg/controller/services.go
+++ b/pkg/controller/services.go
@@ -1076,6 +1076,13 @@ func (cont *AciController) writeApicSvc(key string, service *v1.Service) {
 		p.SetAttr("nodePort", strconv.Itoa(int(port.NodePort)))
 		aobj.AddChild(p)
 	}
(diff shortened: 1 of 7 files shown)


pkg/controller/services.go
```diff
@@ -1076,6 +1076,13 @@ func (cont *AciController) writeApicSvc(key string, service *v1.Service) {
 		p.SetAttr("nodePort", strconv.Itoa(int(port.NodePort)))
 		aobj.AddChild(p)
 	}
+	if service.ObjectMeta.Labels != nil {
```
tomflynn, line 1079: I'm not sure whether this is the right place but we need code somewhere that checks the APIC version and only tries to add vmmInjectedLabel as a child to APICs that support it (5.2.?+ for ns, svcs, 5.0+ for deployment, rs and pods)

tomflynn (changes requested):

tomflynn (approved):

PR #772, opened 2021-04-13
Updated k8s1.19.3 to k8s1.21.0

jayaramsatya:
go version is upated to 1.16 from 1.15


snaiksat (approved):
Tested that build seems to be going through fine after this change (golang 1.16 is used).

PR #776, opened 2021-04-19
Qos Policy is configurable in APIC versions > 5.1

tanyatukade:
- Qos Policy is configurable in APIC versions > 5.1.
- Policy delete should check the Object delete state as well.


pkg/controller/qos.go
```diff
@@ -118,6 +131,13 @@ func (cont *AciController) qosPolicyDelete(qosobj interface{}) {
 }
 
 func (cont *AciController) handleQosPolUpdate(obj interface{}) bool {
+
+	// Ability to configure Qos Policy is available only in APIC versions >= 5.1(x).
+	if apicapi.ApicVersion < "5.1" {
+		cont.log.Error("Cannot create Qos Policy in APIC versions < 5.1(x). Actual APIC version: ",
+			apicapi.ApicVersion)
+	}
```
tomflynn, line 139: Should you return after logging?

tanyatukade: Yes, fixed it

tomflynn (approved):

PR #778, opened 2021-04-19
Fixed  Session Affinity stickiness issue when session affinity config is  removed

jayaramsatya:
This issue is observed when the session affinity config is added and removed with large timeouts.
After removing the session affinity kernel flow is not getting timed-out, due to opflex-agent do not cache the old flows to get the diff out of it. the solution is to set the timeout to be 1 sec. so that flow will be expired.

### DIFF
--- a/pkg/hostagent/services.go
+++ b/pkg/hostagent/services.go
@@ -86,7 +86,10 @@ type opflexClientIPConfig struct {
 }
 
 // Default Session value is 10800(for 3 hours)
-const DefaultSessionAffinityTimer = 10800
+const (
+	DefaultSessionAffinityTimer = 10800
+	TempSessionAffinityTimer    = 1
+)
 
 // Name of the Openshift Service
 const (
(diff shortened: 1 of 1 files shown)


pkg/hostagent/services.go
```diff
@@ -618,9 +621,7 @@ func (sep *serviceEndpoint) SetOpflexService(ofas *opflexService, as *v1.Service
 			} else {
 				sm.ServiceIp = as.Spec.ClusterIP
 			}
-			if as.Spec.SessionAffinityConfig != nil && as.Spec.SessionAffinity == "ClientIP" {
-				sm.SessionAffinity = getSessionAffinity(as.Spec.SessionAffinityConfig)
-			}
+			sm.SessionAffinity = getSessionAffinity(as)
```
abhis2112, line 624: is it not fine as the way it is earlier that ---> if check first and then function call rather than doing it inside the function. In the older way, we can save the function call and push pop of calls on the program stack

pkg/hostagent/services.go
```diff
@@ -618,9 +621,7 @@ func (sep *serviceEndpoint) SetOpflexService(ofas *opflexService, as *v1.Service
 			} else {
 				sm.ServiceIp = as.Spec.ClusterIP
 			}
-			if as.Spec.SessionAffinityConfig != nil && as.Spec.SessionAffinity == "ClientIP" {
-				sm.SessionAffinity = getSessionAffinity(as.Spec.SessionAffinityConfig)
-			}
```
abhis2112, line 623: and writing else here as :
sm.SessionAffinity = &opflexSessionAffinityConfig{ClientIP: opflexClientIPConfig{TimeoutSeconds: TempSessionAffinityTimer}}

pkg/hostagent/services.go
jayaramsatya: No, here we are always creating a 1Sec timer. so no need to have a call

abhis2112: As discussed, the comment was minor and for generic refactoring code as if needed. The comment was majorly focused to save a function call. But the bigger use case was for modularising it in a single place which takes the higher priority. Aprooving

abhis2112 (approved):

PR #785, opened 2021-04-22
Changes for acc-provision operator

apoorva11029:
A new operator is introduced for cluster upgrades and lifecycle management. This requires the acicnioperator pod to read acc-provision configmap and create a AccProvisionInput CR based on the unmarshalled data.
Relevant files changed/added:
1. cmd/acicontainersoperator/main.go: Add a REST client for acc-provision operator
2. pkg/accprovisioninput/: New package with the acc-provision CRD definition used by the acicnioperator to construct the CR
3. pkg/controller/acicontainersoperator.go: Add code to read the acc-provision configmap and create CR

Additionally fix a small bug in the Status field definition for acicnicontainersoperator CRD


pkg/controller/acicontainersoperator.go
```diff
@@ -425,6 +433,64 @@ func (c *Controller) CreateAciContainersOperatorCR() error {
...
+	var err error
+	if err := wait.PollInfinite(time.Second*2, func() (bool, error) {
+		_, er := c.AccProvisionInput_Clientset.AciV1alpha1().AccProvisionInputs(os.Getenv("SYSTEM_NAMESPACE")).Create(context.TODO(), obj, metav1.CreateOptions{})
+		if er != nil {
+			if errors.IsAlreadyExists(er) { //Happens due to etcd timeout
+				log.Info(er)
+				return true, nil
+			}
+			log.Info("Waiting for CRD to get registered to etcd....: ", er)
 			log.Info("Waiting for CRD to get registered to etcd....: ", err)
```
abhis2112, line 494: Duplicate line

anmol372 (commented):
Is vendor/k8s.io/code-generator/generate-groups.sh file supposed to be empty.
Everything else looks good to me.

pkg/controller/acicontainersoperator.go
```diff
@@ -469,6 +535,39 @@ func (c *Controller) Run(stopCh <-chan struct{}) {
 			}
 		}
 	}
+
```
abhis2112, line 538: Do we have to enable the creation of accprovisioninput CR once acicnioperator CR is created or both are independent? If in case acicnioperator CR is not created, do we need to go ahead create the accprovisioninput ?

abhis2112 (changes requested):
Request Feedback

apoorva11029:
> Is vendor/k8s.io/code-generator/generate-groups.sh file supposed to be empty.
> Everything else looks good to me.

Thanks for the review. The file seems empty in the PR view somehow, but it has content- see this https://github.com/noironetworks/aci-containers/blob/6d33abbaa8bd603891249c655985c176f970bee4/vendor/k8s.io/code-generator/generate-groups.sh

pkg/controller/acicontainersoperator.go
apoorva11029: Thanks for pointing it out. 
I have introduced a new variable to ensure that we don't work on the second CR till acicnioperator CR is created https://github.com/noironetworks/aci-containers/blob/f928ebfc418d23bbcecd1236881d07235da3fc5c/pkg/controller/acicontainersoperator.go#L505

snaiksat (changes requested):

pkg/controller/acicontainersoperator.go
```diff
@@ -465,10 +533,45 @@ func (c *Controller) Run(stopCh <-chan struct{}) {
...
 		}
 	}
+	if acicnioperatorsuccess == true {
+		log.Info("Checking if accprovisioninput CR already present")
+		accprovisioninput, err := c.GetAccProvisionInputCR()
+		if err != nil {
+			log.Info("Not Present ..Creating accprovisioninput CR")
+			er := c.CreateAccProvisionInputCR()
+			if er != nil {
+				log.Error(er)
```
snaiksat, line 548: Here and below, wherever we are logging error, is it okay to proceed in spite of error?

cmd/acicontainersoperator/main.go
```diff
@@ -42,6 +43,23 @@ func getOperatorClient() operatorclientset.Interface {
...
+func getAccProvisionClient() accprovisioninputclientset.Interface {
+
+	restconfig, err := restclient.InClusterConfig()
+	if err != nil {
+		return nil
+	}
+
+	AccProvisionClient, err := accprovisioninputclientset.NewForConfig(restconfig)
+	if err != nil {
+		log.Fatalf("Failed to intialize ACC Provision operator client %v", err)
```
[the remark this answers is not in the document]
apoorva11029: The log.Fatalf would make the program exit anyway, right?

pkg/controller/acicontainersoperator.go
apoorva11029: The operator repeatedly tries to create the CR even if there's an error so we can proceed here. @abhis2112 does that sound right?

cmd/acicontainersoperator/main.go
```diff
@@ -42,6 +43,23 @@ func getOperatorClient() operatorclientset.Interface {
 	return OperatorClient
 }
 
+// Create ACC provision Client
+func getAccProvisionClient() accprovisioninputclientset.Interface {
+
+	restconfig, err := restclient.InClusterConfig()
+	if err != nil {
```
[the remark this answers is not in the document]
apoorva11029: Added an error log here

cmd/acicontainersoperator/main.go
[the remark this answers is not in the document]
snaiksat: But then we seem to be returning here:
https://github.com/noironetworks/aci-containers/pull/785/files#diff-025041403c903a35fd3371456b93d4d2b46a01a51ec43b6c64b28b4a68636a14R92

pkg/accprovisioninput/apis/aci.ctrl/v1alpha1/types.go
```diff
@@ -0,0 +1,126 @@
...
+	PBRTrackingNonSNAT bool `json:"pbr_tracking_non_snat,omitempty"`
+	PodSubnetChunkSize int  `json:"pod_subnet_chunk_size,omitempty"`
+}
+
+type RegistrySpec struct {
+	ImagePrefix                    string `json:"image_prefix,omitempty"`
+	ImagePullSecret                string `json:"image_pull_secret,omitempty"`
+	AccProvisionOperatorVersion    string `json:"acc_provision_operator_version,omitempty"`
+	AciContainersControllerVersion string `json:"aci_containers_controller_version,omitempty"`
+	AciContainersHostVersion       string `json:"aci_containers_host_version,omitempty"`
```
snaiksat, line 63: aci_containers_operator_version is missing here...

pkg/accprovisioninput/apis/aci.ctrl/v1alpha1/types.go
```diff
@@ -0,0 +1,126 @@
+package v1alpha1
```
snaiksat, line 1: Few of the configs which I pointed out in the acc-provision patch are also missing here...

cmd/acicontainersoperator/main.go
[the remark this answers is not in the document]
apoorva11029: Not sure if we need the return there, but I can go ahead and add.

pkg/accprovisioninput/apis/aci.ctrl/v1alpha1/types.go
apoorva11029: The new Ansible operator generates the acicnioperator CR spec which currently includes the CNI deployment file (https://github.com/noironetworks/acc-provision/blob/master/provision/acc_provision/templates/aci-containers.yaml) but not the acicnioperator deployment spec. Do we want the Ansible operator to manage the acicnioperator as well?

cmd/acicontainersoperator/main.go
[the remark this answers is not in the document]
snaiksat: I am saying (1) we need to be sure about the behavior (that it fatalf does return) (2) be consistent (i.e. if we dont need a return, we should not have it everywhere)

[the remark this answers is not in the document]
apoorva11029: I see. log.Fatalf is basically a panic so the code would exit- wondering though if we should be panicking in our code like that?

pkg/accprovisioninput/apis/aci.ctrl/v1alpha1/types.go
snaiksat: I am not sure I understand the question, but we have one "operator" which should be able to manage all the components of the ACI CNI. So, when the image tag for the acc-provision-operator container is updated, it should result in all the components getting updated accordingly.

Also, independently, it should be possible to only update the aci-containers-operator container image tag (as can be done with other ACI CNI images listed here).

cmd/acicontainersoperator/main.go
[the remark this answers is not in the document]
apoorva11029: We are currently doing something similar in the rest of our code too https://github.com/noironetworks/aci-containers/blob/c507ef65e885c2880db112cb67913246c70cfd6a/cmd/controller/main.go#L53
Probably by design then...

abhis2112 (approved):

snaiksat (approved):

PR #797, opened 2021-05-08
Code to sync snatglobalinfo and nodeinfos on the controller, logging changes

apoorva11029:
1. The controller now watches the snatglobalinfo object for updates, and verifies the value against nodeinfos- if there's a discrepancy, we queue the nodeinfo to update the snatglobalinfo
2. Modify log level for SNAT code and add new logs

### DIFF
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -19,6 +19,7 @@ import (
 	"fmt"
 	"io/ioutil"
 	"net"
+	"reflect"
 	"strconv"
 	"strings"
 	"sync"
@@ -29,6 +30,7 @@ import (
 	"golang.org/x/time/rate"
 
 	v1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/tools/cache"
 	"k8s.io/client-go/util/workqueue"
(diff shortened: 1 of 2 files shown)


pkg/controller/controller.go
```diff
@@ -630,6 +632,67 @@ func (cont *AciController) Run(stopCh <-chan struct{}) {
 			cont.opflexDeviceDeleted(dn)
 		})
 	go cont.apicConn.Run(stopCh)
+	go cont.enableGlobalSync(stopCh)
+}
+
+func (cont *AciController) enableGlobalSync(stopCh <-chan struct{}) {
+	time.Sleep(time.Minute)
```
snaiksat, line 639: there should be a log here which says that this sync thread has started, otherwise at info level there is no way of knowing that this thread has already started

pkg/controller/controller.go
```diff
@@ -630,6 +632,67 @@ func (cont *AciController) Run(stopCh <-chan struct{}) {
...
+	for {
+		cont.log.Debug("Running code to sync GlobalInfo with Node infos")
```
snaiksat, line 641: "Running code to sync" -> "Syncing "

and we should consider logging this every perhaps every 10th iteration of the loop or something to reduce the number of logs

pkg/controller/controller.go
```diff
@@ -630,6 +632,67 @@ func (cont *AciController) Run(stopCh <-chan struct{}) {
 			cont.opflexDeviceDeleted(dn)
 		})
 	go cont.apicConn.Run(stopCh)
+	go cont.enableGlobalSync(stopCh)
```
snaiksat, line 635: Is there a reason why this doesn't get called after here instead:
https://github.com/noironetworks/aci-containers/blob/master/pkg/controller/environment.go#L290

pkg/controller/controller.go
```diff
@@ -630,6 +632,67 @@ func (cont *AciController) Run(stopCh <-chan struct{}) {
...
+					expectedMap[v] = true
+				}
+				eq := reflect.DeepEqual(expectedMap, policyNames)
+				if !eq {
+					cont.log.Debug("Wrong snatpolicy entry in snatglobalinfo for node: ", nodeName)
+					marked = true
+				}
+			}
+			if marked {
+				cont.log.Info("Nodeinfo and globalinfo out of sync for node: ", nodeName)
```
snaiksat, line 683: It will be helpful to print the snatglobalinfo for this node, and also the nodeinfo for that node so that we can easily tell what is the diff.

pkg/controller/controller.go
```diff
@@ -630,6 +632,67 @@ func (cont *AciController) Run(stopCh <-chan struct{}) {
 			cont.opflexDeviceDeleted(dn)
 		})
 	go cont.apicConn.Run(stopCh)
```
snaiksat, line 634: As a follow up patch we might want to consider running this function based on a config. By default the config is enabled, but we can disable it if required.

snaiksat (approved):
There are some comments here which can be discussed in a follow up patch. For now I am merging this to facilitate testing (since the core functionality in this PR was tested manually using this branch).

PR #799, opened 2021-05-09
Add option to disable period snatglobalinfo sync

apoorva11029:


pkg/controller/controller.go
```diff
@@ -676,6 +683,8 @@ func (cont *AciController) enableGlobalSync(stopCh <-chan struct{}) {
 				eq := reflect.DeepEqual(expectedMap, policyNames)
 				if !eq {
 					cont.log.Debug("Wrong snatpolicy entry in snatglobalinfo for node: ", nodeName)
+					cont.log.Debug("Snatpolicy list according to snatglobalinfo: ", expected[nodeName])
+					cont.log.Debug("Snatpolicy list according to nodeinfo: ", policyNames)
```
snaiksat, line 687: Adding debug logs here unfortunately does not achieve the purpose of providing the Info level instrumentation that I was requesting after L692. If the logging level is Info we will not see the above. I do understand that these variables are probably out of scope at L692, but there should be a way of expanding the scope.

pkg/controller/controller.go
```diff
@@ -632,13 +632,16 @@ func (cont *AciController) Run(stopCh <-chan struct{}) {
...
+		go cont.enableGlobalSync(stopCh)
+	}
 }
 
 func (cont *AciController) enableGlobalSync(stopCh <-chan struct{}) {
 	time.Sleep(time.Minute)
+	cont.log.Info("Go routine to periodically sync globalinfo and nodeinfo started")
 	for {
-		cont.log.Debug("Running code to sync GlobalInfo with Node infos")
+		cont.log.Debug("Syncing GlobalInfo with Node infos")
```
snaiksat, line 644: This is getting logged in ever loop iteration, can we log it every 5th or 10 iteration to reduce the logging noise in the debug level?

pkg/controller/controller.go
apoorva11029: Then I will make the logs from L666-668 and the other two similar logs to info mode, anyway they would only print when there's a discrepancy. Does that sound good?

pkg/controller/controller.go
apoorva11029: Will take of that

pkg/controller/controller.go
snaiksat: Yeah, if that will get logged only in the case of discrepancy (meaning when the current info level log will get logged), then it definitely serves the purpose.

pkg/controller/controller.go
```diff
@@ -632,24 +632,36 @@ func (cont *AciController) Run(stopCh <-chan struct{}) {
 			cont.opflexDeviceDeleted(dn)
 		})
 	go cont.apicConn.Run(stopCh)
-	go cont.enableGlobalSync(stopCh)
 }
 
 func (cont *AciController) enableGlobalSync(stopCh <-chan struct{}) {
```
snaiksat, line 637: This should have ideally been called snatGlobalInfoSync() unless we have plans to expand syncing other resources also in this function. (This can be taken up in a follow-up patch.)

pkg/controller/controller.go
```diff
@@ -658,24 +670,27 @@ func (cont *AciController) enableGlobalSync(stopCh <-chan struct{}) {
 			marked := false
 			policyNames := value.Spec.SnatPolicyNames
 			nodeName := value.ObjectMeta.Name
-			_, ok := expected[nodeName]
+			_, ok := expectedmap[nodeName]
 			if !ok && len(policyNames) > 0 {
-				cont.log.Debug("Missing entry in snatglobalinfo for node: ", nodeName)
+				cont.log.Info("Missing entry in snatglobalinfo for node: ", nodeName)
+				cont.log.Info("No snat policies found in snatglobalinfo")
```
jayaramsatya, line 676: Why many info logs here

jayaramsatya (approved):
LGTM

PR #807, opened 2021-05-17
Epg subscriptions upon EPG update 

Bhavanaashok33:
Added epgdns subscription for epg bindings to a vmm domains on APIC and clear fault instances upon controller restart

### DIFF
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -157,7 +157,8 @@ type AciController struct {
 	// named networkPolicies
 	nmPortNp map[string]bool
 	// cache to look for VRF Epg DNs
-	cachedVRFDns []string
+	cachedVRFDns     []string
+	faultInstCleared bool
 }
 
 type nodeServiceMeta struct {
@@ -630,6 +631,19 @@ func (cont *AciController) Run(stopCh <-chan struct{}) {
 		func(dn string) {
 			cont.opflexDeviceDeleted(dn)
 		})
+
+	cont.apicConn.AddSubscriptionClass("fvAEPg",
+		[]string{"fvAEPg"}, "")
+
+	cont.apicConn.SetSubscriptionHooks("fvAEPg",
+		func(obj apicapi.ApicObject) bool {
+			cont.epgClassChanged(obj)
+			return true
+		},
+		func(dn string) {
+			cont.epgClassDeleted(dn)
+		})
+
 	go cont.apicConn.Run(stopCh)
 }
 
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -157,9 +157,9 @@ type AciController struct {
 	ctrPortNameCache map[string]*ctrPortNameEntry
 	// named networkPolicies
 	nmPortNp map[string]bool
-	// cache to look for VRF Epg DNs
-	cachedVRFDns     []string
-	faultInstCleared bool
+	// cache to look for Epg DNs which are bound to Vmm domain
+	cachedEpgDns             []string
+	vmmClusterFaultSupported bool
 }
 
 type nodeServiceMeta struct {
@@ -527,6 +527,9 @@ func (cont *AciController) Run(stopCh <-chan struct{}) {
 		} else {
 			cont.apicConn.SnatPbrFltrChain = false
 		}
+		if version >= "5.2" {
+			cont.vmmClusterFaultSupported = true
+		}
 	} else { // For unit-tests
 		cont.apicConn.SnatPbrFltrChain = true
 	}
@@ -542,6 +545,26 @@ func (cont *AciController) Run(stopCh <-chan struct{}) {
 			cont.log.Error("Pod/NodeBDs and AciL3Out VRF association is incorrect")
 			panic(err)
 		}
+	}
+
+	if len(cont.config.ApicHosts) != 0 && cont.vmmClusterFaultSupported {
+		//Clear fault instances when the controller starts
+		cont.clearFaultInstances()
+		//Subscribe for vmmEpPD for a given domain
+		var tnTargetFilterEpg string
+		tnTargetFilterEpg += fmt.Sprintf("uni/vmmp-%s/dom-%s", cont.vmmDomainProvider(), cont.config.AciVmmDomain)
+		subnetTargetFilterEpg := fmt.Sprintf("and(wcard(vmmEpPD.dn,\"%s\"))", tnTargetFilterEpg)
+		cont.apicConn.AddSubscriptionClass("vmmEpPD",
+			[]string{"vmmEpPD"}, subnetTargetFilterEpg)
+		cont.apicConn.SetSubscriptionHooks("vmmEpPD",
+			func(obj apicapi.ApicObject) bool {
+				cont.vmmEpPDChanged(obj)
+				return true
+			},
+			func(dn string) {
+				cont.vmmEpPDDeleted(dn)
+			})
+
 	}
 
 	cont.initStaticObjs()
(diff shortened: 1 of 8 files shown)


Bhavanaashok33:
Next PR will include some UTs to test EPG subscription upon epg add , epg delete on APIC and clean up fault instances upon controller restart. I will be creating a separate PR for the UTs if thats fine.

Bhavanaashok33:
Please go ahead with the review!

pkg/controller/controller.go
```diff
@@ -543,6 +547,26 @@ func (cont *AciController) Run(stopCh <-chan struct{}) {
 		}
 	}
 
+	if len(cont.config.ApicHosts) != 0 && cont.vmmClusterFaultSupported {
+		//Clear fault instances when the controller starts
+		cont.clearFaultInstances()
+		//Subscribe for vmmEpPD for a given domain
+		var tnTargetFilterEpg string
+		tnTargetFilterEpg += fmt.Sprintf("uni/vmmp-Kubernetes/dom-%s", cont.config.AciVmmDomain)
```
tomflynn, line 555: vmmp-Kubernetes can't be hardcoded. The vendor needs to come from config as it could be OpenShift instead

pkg/controller/epgcache.go
```diff
@@ -0,0 +1,128 @@
...
+func (cont *AciController) clearFaultInstances() {
+	args := []string{
+		"order-by=vmmClusterFaultInfo.modTs|desc",
+	}
+	faultUri := fmt.Sprintf("/api/node/class/vmmClusterFaultInfo.json?%s", strings.Join(args, "&"))
+	apicresp, err := cont.apicConn.GetApicResponse(faultUri)
+	if err != nil {
+		return
+	}
+	subString := fmt.Sprintf("comp/prov-Kubernetes/ctrlr-[%s]-%s/injcont/info/", cont.config.AciVmmDomain, cont.config.AciVmmDomain)
```
tomflynn, line 37: again can't hardcode prov-Kubernetes

pkg/controller/epgcache.go
```diff
@@ -0,0 +1,128 @@
...
+	"fmt"
+	"sort"
+	"strings"
+)
+
+func (cont *AciController) clearFaultInstances() {
+	args := []string{
+		"order-by=vmmClusterFaultInfo.modTs|desc",
+	}
+	faultUri := fmt.Sprintf("/api/node/class/vmmClusterFaultInfo.json?%s", strings.Join(args, "&"))
```
tomflynn, line 32: Is there a way to query just for vmmClusterFaultInfo instances under this VMM instead of retrieving all and skipping the ones for other VMMs?

pkg/controller/epgcache.go
```diff
@@ -0,0 +1,128 @@
...
+				}
+				cont.indexMutex.Unlock()
+			}
+		}
+	}
+	return
+}
+
+func (cont *AciController) getSubString(dn string) string {
+	delimiterString := fmt.Sprintf("uni/vmmp-Kubernetes/dom-%s/eppd-[", cont.config.AciVmmDomain)
```
tomflynn, line 55: vmmp-Kubernetes hardcoded

pkg/controller/epgcache.go
```diff
@@ -0,0 +1,128 @@
...
+	epgDn := strings.Split(dn, delimiterString)[1]
+	epgDn = strings.Replace(epgDn, "]", "", -1)
+	return epgDn
+}
+
+func (cont *AciController) epgClassChanged(obj apicapi.ApicObject) {
+	dn := obj.GetAttrStr("dn")
+	epgDn := cont.getSubString(dn)
+	cont.indexMutex.Lock()
+	if ok := !cont.contains(cont.cachedVRFDns, epgDn); ok {
```
tomflynn, line 65: It's kind of strange that we're storing EPG DNs in 'cachedVRFDns'. Does it make sense to have a different var for this? Same comment applies to func names below

pkg/controller/pods.go
```diff
@@ -112,30 +112,37 @@ func (cont *AciController) checkIfEpgExistPod(pod *v1.Pod) {
...
+
+			aPrObj := apicapi.NewVmmInjectedClusterInfo(cont.vmmDomainProvider(),
+				cont.config.AciVmmDomain, cont.config.AciVmmController)
+			aPrObjDn := aPrObj.GetDn()
+			aObj := apicapi.NewVmmClusterFaultInfo(aPrObjDn, faultcode)
+			aObj.SetAttr("faultDesc", desc)
+			aObj.SetAttr("faultCode", faultcode)
+			aObj.SetAttr("faultSeverity", severity)
+
+			if faultCode == 11 {
```
tomflynn, line 139: Can 11 be a constant somewhere?

tomflynn (changes requested):

Bhavanaashok33:
Working on the review comments..

pkg/controller/epgcache.go
Bhavanaashok33: Yes this is achievable by applying the query filters. I have incorporated this change

Bhavanaashok33:
I have addressed all the review comments. Please take a look at it..

pkg/controller/epgcache.go
```diff
@@ -0,0 +1,130 @@
...
+				cont.apicConn.DeleteDn(dn)
+				cont.log.Debug("Clearing existing Fault Instance: ", dn)
+				cont.indexMutex.Unlock()
+			}
+		}
+	}
+	return
+}
+
+func (cont *AciController) getSubString(dn string) string {
```
tomflynn, line 56: can you rename to something like getEpgFromEppd or something to make it clearer what the function is doing?

pkg/controller/epgcache.go
```diff
@@ -0,0 +1,130 @@
...
+
+func (cont *AciController) epgClassChanged(obj apicapi.ApicObject) {
+	dn := obj.GetAttrStr("dn")
+	epgDn := cont.getSubString(dn)
+	cont.indexMutex.Lock()
+	if ok := !cont.contains(cont.cachedEpgDns, epgDn); ok {
+		cont.cachedEpgDns = append(cont.cachedEpgDns, epgDn)
+	}
+	cont.indexMutex.Unlock()
+	sort.Strings(cont.cachedEpgDns)
```
tomflynn, line 71: Shouldn't the sort be protected by the lock as well?

pkg/controller/epgcache.go
```diff
@@ -0,0 +1,130 @@
...
+func (cont *AciController) removeSlice(s []string, searchterm string) {
+	i := sort.SearchStrings(s, searchterm)
+	if i < len(s) && s[i] == searchterm {
+		s = append(s[:i], s[i+1:]...)
+	}
+	return
+}
+
+func (cont *AciController) getCachedEpgDns() []string {
+	return cont.cachedEpgDns
```
tomflynn, line 92: should you be returning a copy here or are all callers expected to use the lock indexMutex

actually the only place I see this used is another member function...couldn't you just acquire the lock and reference cont.cachedEpgDns from there as well?

pkg/controller/epgcache.go
Bhavanaashok33: Okay will do

pkg/controller/epgcache.go
Bhavanaashok33: In older code structure, one of the member function asked for a getter to access the cache. Now that I modified the code a bit, I retained it. In case if needed to access the list of epg bound to vmm domain from a diff function, we can use this getter. Sure, I can lock it while referencing this.

Bhavanaashok33: Well I will take that back. We don't need a getter method since this entity belongs to controller struct. I can access it directly by acquiring a lock from anywhere.

tomflynn (approved):

PR #812, opened 2021-05-25
WIP: makes wait_for_network duration configurable

anmol372:
WIP

### DIFF
--- a/cmd/opflexagentcni/main.go
+++ b/cmd/opflexagentcni/main.go
@@ -57,11 +57,12 @@ type K8SArgs struct {
 
 type NetConf struct {
 	types.NetConf
-	LogLevel       string `json:"log-level,omitempty"`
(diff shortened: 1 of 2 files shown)


cmd/opflexagentcni/main.go
```diff
@@ -99,6 +100,10 @@ func loadConf(args *skel.CmdArgs) (*NetConf, *K8SArgs, string, error) {
 	id := args.ContainerID
 	log.Debug("Args: ", k8sArgs)
 
+	if n.WaitForNetwork && n.WaitForNetworkDuration == 0 {
+		n.WaitForNetworkDuration = 210
+	}
+
```
snaiksat, line 106: perhaps we can also add a debug log here which logs n.WaitForNetworkDuration final value

PR #816, opened 2021-05-27
Add EPRegistry check back in place for podif CRD

tanyatukade:
- The check was removed to enable ERSPAN feature in on-prem case. This caused issues with the way podifs are processed at scale. This commit reverts the change and puts the check back in place. 

- To enable ERSPAN feature, we need to derive the config from acc-provision input file. An example of the config is below:
  
```
kube_config:
  …
  ep_registry: k8s
```


snaiksat (approved):

tomflynn:
A corresponding change will be needed in acc-provision as this param is not being added to the host-agent config anymore for overlay setups

tomflynn (approved):

PR #823, opened 2021-06-14
Polulate the dnsName for the service class vmmInjectedSvc

Bhavanaashok33:
This PR populates the the dnsName for the service vmmInjectedSvc for different service types. 
In case of aws, for a loadbalancer service,  the dns name is allocated by aws itself. But for NodePort, ClusterIP, ExternalName services, FQDN by default would be <service-name>.<namespace>.svc.cluster.local

Verified the DN posting on APIC 

Verified the dnsName assigned by loadbalancer service on aws 

Corresponding issue: https://github.com/noironetworks/support/issues/1380

### DIFF
--- a/pkg/controller/config.go
+++ b/pkg/controller/config.go
@@ -203,6 +203,9 @@ type ControllerConfig struct {
 	CSRTunnelIDBase int `json:"csr-tunnel-id-base,omitempty"`
 	// enable EndpointSlice
 	EnabledEndpointSlice bool `json:"enable_endpointslice,omitempty"`
+
+	// Cluster Flavour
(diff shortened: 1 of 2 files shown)


snaiksat:
Per the issue noironetworks/support#1380 is a check required for APIC > 5.1?

Bhavanaashok33:
@snaiksat  Thanks for catching that. Will add  Apic version check > 5.1 in the next commit.

Bhavanaashok33:
@snaiksat, I have addressed the review comments. Please review it when you can! Thanks

snaiksat (approved):

PR #828, opened 2021-06-28
Fix bug in SNAT policy computation logic

apoorva11029:
Currently we compute IP and port allocations for every policy when there's a change in any snat policy. This causes us to allocate additional port ranges to a node and policy pair. To avoid this, we want to skip computing if the policy is already applied on that node.

Also added more logging on INFO and DEBUG levels for SNAT

### DIFF
--- a/pkg/controller/snatglobalinfo.go
+++ b/pkg/controller/snatglobalinfo.go
@@ -248,8 +248,14 @@ func (cont *AciController) handleSnatNodeInfo(nodeinfo *nodeinfo.NodeInfo) bool 
 			}
 			cont.log.Debug("SnatPolicy Name: ", name)
 			if len(snatpolicy.SnatIp) != 0 {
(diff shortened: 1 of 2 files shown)


pkg/controller/snatglobalinfo.go
```diff
@@ -248,6 +248,10 @@ func (cont *AciController) handleSnatNodeInfo(nodeinfo *nodeinfo.NodeInfo) bool
 			}
 			cont.log.Debug("SnatPolicy Name: ", name)
 			if len(snatpolicy.SnatIp) != 0 {
+				nodeSNATEntryFound := cont.checkIfPolicyApplied(nodename, name, snatpolicy.ExpandedSnatIps)
+				if nodeSNATEntryFound {
```
snaiksat, line 252: Would it make sense to log here and in L380?

apoorva11029: Added debug log

snaiksat (approved):

PR #829, opened 2021-06-29
Supporting new flavors in aci-operator for OCP-4.7

abhis2112:


snaiksat:
Can we preemptively make the changes for the ESX flavor as well so that we don't have to do another round of PRs?

abhis2112:
Supporting new flavors in aci-operator for OCP-4.7
    1. openshift-4.7-openstack
    2. openshift-4.6-esx
    3. openshift-4.7-esx

> Can we preemptively make the changes for the ESX flavor as well so that we don't have to do another round of PRs?

resolved

snaiksat:
For ESX, we also need to update the following, right?

https://github.com/noironetworks/aci-containers/blob/master/pkg/hostagent/services.go#L114

abhis2112:
> For ESX, we also need to update the following, right?
> 
> https://github.com/noironetworks/aci-containers/blob/master/pkg/hostagent/services.go#L114

Sure Sumit, Let me add here too in this commit. Was going to ask about it whether we mention it in a separate commit just to record it as ingress change rather than flavor type commit.

snaiksat (approved):

PR #833, opened 2021-07-13
Introduce NodePodIF CRD to handle scale related issues

tanyatukade:
- The original PodIF CRD is left intact and will only be used in the pure overlay flavor.

- Refer: ACI-CNI: Optimize PodIF code to reduce processing overhead and decrease delaying the programming of datapath for new pods support#1539

- The new NodePodIF CRD has a list of podifs for all pods on each node. Hence each node will have only 1 NodePodIF CR. Therefore the API requests happen per node rather than per pod.

- A config knob is added in the hostagent config to enable nodepodif. By default it is not enabled. Refer: https://github.com/noironetworks/acc-provision/pull/540

- The consumers of podif such as the ERSPAN feature will use nodepodifs instead of podifs.

- Moved PodIF attributes from Status to Spec field.

- Corresponding Acc-Provision PR: https://github.com/noironetworks/acc-provision/pull/535


pkg/util/nodepodif.go
```diff
@@ -0,0 +1,30 @@
+// Copyright 2019 Cisco Systems, Inc.
+//
```
snaiksat, line 2: nit: 2021

snaiksat:
Changes look good, waiting on the UTs.

tanyatukade:
> Changes look good, waiting on the UTs.

Updated the PR with UTs

apoorva11029 (approved):

PR #845, opened 2021-08-09
Add locks to secure SNAT go routines in hostagent

apoorva11029:


pkg/hostagent/pod_relatives.go
```diff
@@ -323,6 +327,8 @@ func deploymentLogger(log *logrus.Logger, dep *appsv1.Deployment) *logrus.Entry
 
 func (agent *HostAgent) deploymentAdded(obj interface{}) {
 	agent.log.Infof("deploymentAdded => ")
+	agent.indexMutex.Lock()
+	defer agent.indexMutex.Unlock()
```
snaiksat, line 331: Is the lock not required for the call from L362 as well?

apoorva11029: Yes, missed it

snaiksat (approved):

PR #868, opened 2021-08-27
Add mutex lock when writing to CachedSubnetDns and reading from Servi…

tanyatukade:
…ceEndpoint

Refer: https://github.com/noironetworks/support/issues/1612

### DIFF
--- a/pkg/controller/services.go
+++ b/pkg/controller/services.go
@@ -314,7 +314,7 @@ func apicRedirectDst(rpDn string, ip string, mac string,
 	return dst
 }
 
-func apicRedirectPol(name string, tenantName string, nodes []string,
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -35,7 +35,7 @@ DIST_FILE=aci-containers.tgz
 
 DOCKER_HUB_ID ?= noiro
 DOCKER_TAG ?=
-BUILD_CMD ?= go build -v -trimpath
+BUILD_CMD ?= go build -v -trimpath -race
```
snaiksat, line 38: we can add a new build command option here, but the existing one should not be modified...

snaiksat (approved):

PR #870, opened 2021-08-29
hack_1

anmol372:


pkg/hostagent/setup.go
```diff
@@ -281,23 +289,26 @@ func (agent *HostAgent) configureContainerIfaces(metadata *md.ContainerMetadata)
...
 		}
+		//Can we check if pod still exists here, and gracefully fail?
 		for _, ip := range iface.IPs {
 			//There are 4 cases: IPv4-only, IPv6-only, dual stack with either IPv4 or IPv6 as the first address.
 			//We are guaranteed to derive the MAC address from IPv4 if it is assigned
 			if ip.Address.IP != nil && ip.Address.IP.To4() != nil {
 				iface.HostVethName, iface.Mac, err =
 					runSetupVeth(iface.Sandbox, iface.Name, mtu, ip.Address.IP)
 				if err != nil {
+					deallocIP(iface)
```
snaiksat, line 311: I think it will help to log here at info level that "runSetupVeth" has returned an (we can log the error as well) and that we are returning an error.

pkg/hostagent/setup.go
```diff
@@ -306,21 +317,26 @@ func (agent *HostAgent) configureContainerIfaces(metadata *md.ContainerMetadata)
 		}
 		// if no mac is assigned, set it to the default Mac.
 		if len(iface.Mac) == 0 {
+			logger.Infof("No Mac assigned, assigning default")
 			iface.HostVethName, iface.Mac, err =
 				runSetupVeth(iface.Sandbox, iface.Name, agent.config.InterfaceMtu, nil)
 			if err != nil {
+				deallocIP(iface)
```
snaiksat, line 324: I think it will help to log here at info level that "runSetupVeth" has returned an (we can log the error as well) and that we are returning an error.

pkg/hostagent/setup.go
```diff
@@ -306,21 +317,26 @@ func (agent *HostAgent) configureContainerIfaces(metadata *md.ContainerMetadata)
...
 				return nil, err
 			}
 		}
 
+		if len(iface.HostVethName) == 0 || len(iface.Mac) == 0 {
+			deallocIP(iface)
+			return nil, fmt.Errorf("Unable to Configure Container Interface")
```
snaiksat, line 331: I think it will also help to log the values of len(iface.HostVethName) and len(iface.Mac) here.

pkg/hostagent/setup.go
```diff
@@ -306,21 +317,26 @@ func (agent *HostAgent) configureContainerIfaces(metadata *md.ContainerMetadata)
...
+
 		agent.addToResult(iface, ifaceind, result)
 
 		logger.Debug("Configuring network for ", iface.Name, ": ", *result)
 		err = runSetupNetwork(iface.Sandbox, iface.Name, result)
 		if err != nil {
-			agent.ipamMutex.Lock()
-			agent.deallocateIpsLocked(iface)
-			agent.ipamMutex.Unlock()
+			deallocIP(iface)
```
snaiksat, line 339: I think it will help to error log here that "runSetupNetwork" failed (perhaps also the err) and that we are returning.

pkg/hostagent/ipam.go
```diff
@@ -192,6 +192,7 @@ func (agent *HostAgent) deallocateMdIps(md *metadata.ContainerMetadata) {
 				continue
 			}
 
+			agent.log.Infof("deallocating IP: %v", ip.Address.IP)
```
snaiksat, line 195: It might be helpful to make logging of the deallocateMdIps function consistent with this function (i.e., the before and after logs)

snaiksat: Also, why is it that in this case iface.IPs was not being set to nil as is being in L178?

anmol372: Afaik, we deallocate with  deallocateMdIps when the containers metadata exists, the calling function here takes care of clearing stale metadata, (ContainerIfaceMd, epMetadata)

On the other hand when the container metadata doesn't exist, we are setting it to nil manually, i.e. the function deallocateIpsLocked is called when for some reason an IP may have been assigned but the container fails to come up for other reasons.

pkg/hostagent/setup.go
anmol372: We already log the error from runSetupVeth in the calling function https://github.com/noironetworks/aci-containers/blob/87ece9a8e4598995204b0041cbb7fe3de1df561e/pkg/hostagent/eprpc.go#L89.

Should we add another?

PR #873, opened 2021-09-01
Ipam leak master

anmol372:
Not the best way I could think of to modify the integration test. 
The problem I'm facing here is that setupVeth makes some system level calls and expects to read from a network namespace file. So I think an approach could be to mock a netns file for the integ-test.


pkg/hostagent/integ_test.go
```diff
@@ -342,6 +344,74 @@ func mkPod(uuid string, namespace string, name string,
 	}
 }
 
+func TestIPAMadd(t *testing.T) {
```
anmol372, line 347: Please ignore this test, it's only for my experimentation

snaiksat (approved):

PR #875, opened 2021-09-02
collect versions and more details in the cluster-report

vlella:


cmd/acikubectl/cmd/debug.go
```diff
@@ -483,6 +495,29 @@ func accLogCmdArgs(systemNamespace string) []string {
 		"-c", "aci-containers-controller"}
 }
 
+func acikubectlversion(systemNamespace string) []string {
```
snaiksat, line 498: This does not look like what are trying to get for the acikubectl version. The expectation was for example this:

     noiro@fab14-compute-1:~$ acikubectl version
     GitCommit: 20c082416c0aefb73b9af4b275ac9fd9f74b722c
     BuildTime: 08-31-2021.08:46:01.UTC

snaiksat (changes requested):

cmd/acikubectl/cmd/debug.go
vlella: if you check the output on fab-14 at this path /home/noiro/cluster-report/logs/versions and see the newly generated acikubectl_version.log in the cluster report it has the same output as acikubectl version command above.

it's already implemented here - https://github.com/noironetworks/aci-containers/blob/master/cmd/acikubectl/cmd/version.go#L28
I just took the logic and modified in a way that it works for the codeflow in debug.go

I tried initially to implement kubectl aci version (i.e use of kubectl for acikubectl commands) but it was throwing error when generating cluster report stating - (Error: flags cannot be placed before plugin name: --kubeconfig) which didn't throw in normal case as it's been handled via acc-provision. So to run kubectl aci commands we need to setup few things before which is tough to implement via debug.go that's why took the logic from version.go and tweaked it for my use case. If required I can explain in detail as well later.

snaiksat: > it has the same output as acikubectl version command above

That is because acikubectl (the tool) and aci-containers-controller (the container) are being built from the same code base. That is incidental to this setup and is what we are trying to validate. That misses the requirement here, which is to get the output of:

        acikubectl version

as opposed to what you have now, which is:

       kubectl exec -c aci-containers-controller -n aci-containers-system <aci-containers-controller-pod-name> -- /bin/sh -c aci-containers-controller --version

snaiksat: On further thinking, since internally acikubectl version is getting the value of aci-containers-controller to display the version, I think its pointless to get the aci-containers-controller version, since we already have that. Ideally we should have had that separately in the acikubectl code, but that is a separate and bigger change. Let's drop this requirement for now.

vlella: okay sure.

snaiksat (approved):

PR #877, opened 2021-09-03
Revise Periodic Snat Global Cache sync test to handle port ranges in …

tanyatukade:
…any order

### DIFF
--- a/pkg/controller/snatglobalinfo_test.go
+++ b/pkg/controller/snatglobalinfo_test.go
@@ -128,17 +128,25 @@ var nodeTests = []nodedata{
 }
 
 func snatWait(t *testing.T, desc string, expected map[string]snatglobalinfo.GlobalInfo,
-	actual map[string]*snatglobalinfo.GlobalInfo) {
+	actual map[string]*snatglobalinfo.GlobalInfo, portRangeSet bool) {
 	tu.WaitFor(t, desc, 100*time.Millisecond, func(last bool) (bool, error) {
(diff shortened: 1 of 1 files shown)


snaiksat (approved):

pkg/controller/snatglobalinfo_test.go
```diff
@@ -128,17 +128,25 @@ var nodeTests = []nodedata{
...
 		for key, v := range expected {
 			val, ok := actual[key]
 			if ok {
-				result := tu.WaitEqual(t, last, v.SnatIp, val.SnatIp, "snatIp does not match") &&
-					tu.WaitEqual(t, last,
-						v.PortRanges[0], val.PortRanges[0], "Portrange does not match") &&
-					tu.WaitEqual(t, last, v.SnatPolicyName, val.SnatPolicyName, "snatPolicyName does not match")
-				if !result {
-					return false, nil
+				if portRangeSet {
```
apoorva11029, line 136: Instead of the two big if conditions we could do thing:
```
result := tu.WaitEqual(t, last, v.SnatIp, val.SnatIp, "snatIp does not match") &&
						tu.WaitEqual(t, last, v.SnatPolicyName, val.SnatPolicyName, "snatPolicyName does not match")
if portRangeSet {
   result = result && tu.WaitEqual(t, last,
							v.PortRanges[0], val.PortRanges[0], "Portrange does not match")
}					
					
```
Does that make sense?

PR #880, opened 2021-09-09
Introduce new Snat Policy Cache RW lock to eliminate data race

tanyatukade:
- Changed the name of snatMutex -> snatPolicyLabelMutex to avoid confusion with the new lock
- This patch eliminates the data race found when accessing snatPolicyCache in hostagent


tanyatukade:
Still working on fixes a few other data races that were revealed by this fix

PR #887, opened 2021-09-10
Introduce new Snat Policy Cache RW lock to eliminate data race

tanyatukade:
- Changed the name of snatMutex -> snatPolicyLabelMutex to avoid confusion with the new lock
- This patch eliminates the data race found when accessing snatPolicyCache in hostagent

### DIFF
--- a/pkg/hostagent/agent.go
+++ b/pkg/hostagent/agent.go
@@ -45,9 +45,10 @@ type HostAgent struct {
 	config *HostAgentConfig
 	env    Environment
 
-	indexMutex sync.Mutex
-	ipamMutex  sync.Mutex
-	snatMutex  sync.RWMutex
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -35,7 +35,7 @@ DIST_FILE=aci-containers.tgz
 
 DOCKER_HUB_ID ?= noiro
 DOCKER_TAG ?=
-BUILD_CMD ?= go build -v -trimpath
+BUILD_CMD ?= go build -v -trimpath -race
```
snaiksat, line 38: before merging we would need to revert this and L50

tanyatukade: sure, testing is in process

snaiksat (approved):

snaiksat (approved):

PR #892, opened 2021-09-15
Adding aci-operator-config config-map watcher to ACI containers operator 

abhis2112:
1. The Watcher watches the aci-operator-config config-map
2. Change in config-map will handle the respective change in acicontainersoperator CR.
3. This in turn apply the latest config to the cluster.
4. This case will handle the update done by acc-provision-operator during the update of aci-operator-configmap
   and passing the change to acicontaineroperator CR.

### DIFF
--- a/pkg/controller/acicontainersoperator.go
+++ b/pkg/controller/acicontainersoperator.go
@@ -36,6 +36,7 @@ import (
 	routesClientset "github.com/openshift/client-go/route/clientset/versioned"
 	log "github.com/sirupsen/logrus"
 	appsv1 "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
 	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -73,11 +74,13 @@ type Controller struct {
 	Daemonset_Queue             workqueue.RateLimitingInterface
 	Node_Queue                  workqueue.RateLimitingInterface
 	Route_Queue                 workqueue.RateLimitingInterface
+	Config_Map_Queue            workqueue.RateLimitingInterface
 	Informer_Operator           cache.SharedIndexInformer
 	Informer_Deployment         cache.SharedIndexInformer
 	Informer_Daemonset          cache.SharedIndexInformer
 	Informer_Node               cache.SharedIndexInformer
 	Informer_Route              cache.SharedIndexInformer
+	Informer_Config             cache.SharedIndexInformer
 	Resources                   AciResources
 	DnsOperatorClient           client.Client             // This client is specific dnsopenshift operator
 	RoutesClient                routesClientset.Interface // This client is specific routes openshift operator
@@ -109,6 +112,9 @@ const aciContainersController = "aci-containers-controller"
 const aciContainersHostDaemonset = "aci-containers-host"
 const aciContainersOvsDaemonset = "aci-containers-openvswitch"
 
+var Aci_operator_config_path = "/usr/local/etc/aci-containers/aci-operator.conf"
+var Acc_provision_config_path = "/usr/local/etc/acc-provision/acc-provision-operator.conf"
+
 func NewAciContainersOperator(
 	acicnioperatorclient operatorclientset.Interface,
 	accprovisioninputclient accprovisioninputclientset.Interface,
@@ -120,6 +126,7 @@ func NewAciContainersOperator(
 	daemonset_queue := workqueue.NewRateLimitingQueue(workqueue.DefaultControllerRateLimiter())
 	node_queue := workqueue.NewRateLimitingQueue(workqueue.DefaultControllerRateLimiter())
 	route_queue := workqueue.NewRateLimitingQueue(workqueue.DefaultControllerRateLimiter())
+	config_map_queue := workqueue.NewRateLimitingQueue(workqueue.DefaultControllerRateLimiter())
 
 	log.Info("Intializing Informer")
 
@@ -177,6 +184,21 @@ func NewAciContainersOperator(
 		0,
 		cache.Indexers{},
 	)
+
+	config_map_informer := cache.NewSharedIndexInformer(
+		&cache.ListWatch{
+			ListFunc: func(options metav1.ListOptions) (runtime.Object, error) {
+				return k8sclient.CoreV1().ConfigMaps(os.Getenv("SYSTEM_NAMESPACE")).List(context.TODO(), options)
+			},
+			WatchFunc: func(options metav1.ListOptions) (watch.Interface, error) {
+				return k8sclient.CoreV1().ConfigMaps(os.Getenv("SYSTEM_NAMESPACE")).Watch(context.TODO(), options)
+			},
+		},
+		&v1.ConfigMap{},
+		0,
+		cache.Indexers{},
+	)
+
 	var routesClient routesClientset.Interface
 	var route_informer cache.SharedIndexInformer
 	flavor := os.Getenv("ACC_PROVISION_FLAVOR")
@@ -221,11 +243,13 @@ func NewAciContainersOperator(
 		Informer_Daemonset:          aci_daemonset_informer,
 		Informer_Node:               node_informer,
 		Informer_Route:              route_informer,
+		Informer_Config:             config_map_informer,
 		Operator_Queue:              operator_queue,
 		Deployment_Queue:            deployment_queue,
 		Daemonset_Queue:             daemonset_queue,
 		Node_Queue:                  node_queue,
 		Route_Queue:                 route_queue,
+		Config_Map_Queue:            config_map_queue,
 		Resources:                   AciResources{},
 		DnsOperatorClient:           nil,
 		RoutesClient:                routesClient,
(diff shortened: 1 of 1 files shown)


pkg/controller/acicontainersoperator.go
```diff
@@ -329,6 +350,35 @@ func NewAciContainersOperator(
 			}
 		},
 	})
+
+	config_map_informer.AddEventHandler(cache.ResourceEventHandlerFuncs{
+		AddFunc: func(obj interface{}) {
+			key, err := cache.MetaNamespaceKeyFunc(obj)
+			log.Debug("The ConfigMap key: ", key)
```
snaiksat, line 357: Assuming this event does not happen often, we should probably log this at Info level. Also, can you make the log a little more descriptive. One should not have to go back and look in the code to understand the context of the log statement.

pkg/controller/acicontainersoperator.go
```diff
@@ -329,6 +350,35 @@ func NewAciContainersOperator(
...
+			current_Obj := obj.(*corev1.ConfigMap)
+			if current_Obj.Name == "aci-operator-config" {
+				if err == nil {
+					config_map_queue.Add(key)
+				}
+			}
+
+		},
+		UpdateFunc: func(prevObj, currentObj interface{}) {
+			log.Debug("In UpdateFunc for ConfigMap")
```
snaiksat, line 367: Info level?

pkg/controller/acicontainersoperator.go
```diff
@@ -329,6 +350,35 @@ func NewAciContainersOperator(
...
+			current_Obj := currentObj.(*corev1.ConfigMap)
+			if current_Obj.Name == "aci-operator-config" {
+				controller.handleConfigMapUpdate(prevObj, currentObj, config_map_queue)
+			}
+		},
+		DeleteFunc: func(obj interface{}) {
+			key, err := cache.MetaNamespaceKeyFunc(obj)
+			log.Debug("Deleted ConfigMap key is :", key)
```
snaiksat, line 375: Info level?

pkg/controller/acicontainersoperator.go
```diff
@@ -383,6 +433,73 @@ func (c *Controller) handledaemonsetUpdate(oldobj interface{}, newobj interface{
...
+	old_cm := oldobj.(*corev1.ConfigMap)
+	new_cm := newobj.(*corev1.ConfigMap)
+
+	if !reflect.DeepEqual(old_cm.Data, new_cm.Data) {
+		key, err := cache.MetaNamespaceKeyFunc(newobj)
+		if err == nil {
+			queue.Add(key)
+		}
+	} else {
+		log.Info("Config-Map is intact for ", new_cm.Name)
```
snaiksat, line 494: intact -> unchanged

pkg/controller/acicontainersoperator.go
```diff
@@ -383,6 +433,73 @@ func (c *Controller) handledaemonsetUpdate(oldobj interface{}, newobj interface{
...
+	}
+}
+
+func (c *Controller) handleConfigMapDelete(obj interface{}) bool {
+	log.Infof("aci-containers-operator config-map Deleted")
```
snaiksat, line 499: nit: but why is "D" upper case here??

pkg/controller/acicontainersoperator.go
```diff
@@ -413,7 +546,7 @@ func (c *Controller) CreateAciContainersOperatorObj() *operators.AciContainersOp
 
 func (c *Controller) CreateAciContainersOperatorCR() error {
 	log.Info("Reading the aci-operator-config Config Map providing CR")
-	rawSpec, err := c.ReadConfigMap("aci-containers/aci-operator.conf")
+	rawSpec, err := c.ReadConfigMap("/usr/local/etc/aci-containers/aci-operator.conf")
```
snaiksat, line 549: Here and elsewhere below - why is this change needed? Making the file path absolute would mean that we have to be really sure that it remains the same in all situations. Would that be the case? And even if that were to be the case, this prefix should probably be pre-defined somewhere else and used here.

pkg/controller/acicontainersoperator.go
```diff
@@ -401,6 +518,22 @@ func (c *Controller) ReadConfigMap(field string) ([]byte, error) {
...
+	if err != nil{
+		log.Error(err)
+		return err
+	}
+	err = ioutil.WriteFile( field, data_byte, 0777)
+	if err != nil {
+		log.Error(err)
+		return err
+	}
+	log.Debug("Config-Map is:  ", (string(data_byte)))
```
snaiksat, line 533: I think we should put an info level log statement stating that we are writing the config-map (name?). The actual contents of the map can be at debug level, but if the config is base-encoded perhaps there is no need to log it.

snaiksat (changes requested):

snaiksat:
Also, need to run gofmt...

pkg/controller/acicontainersoperator.go
abhis2112: Yes the config-map is big base encoded log.  I will change the debug to info level with config map name

pkg/controller/acicontainersoperator.go
```diff
@@ -329,6 +353,35 @@ func NewAciContainersOperator(
 			}
 		},
 	})
+
+	config_map_informer.AddEventHandler(cache.ResourceEventHandlerFuncs{
+		AddFunc: func(obj interface{}) {
+			key, err := cache.MetaNamespaceKeyFunc(obj)
+			log.Info("In AddFunc for Config-Map with key: ", key)
```
snaiksat, line 360: can we put the name of the configmap in the log message? If someone reads this log message in the log, he has to go back to code and check which configmap we are talking about...

pkg/controller/acicontainersoperator.go
```diff
@@ -329,6 +353,35 @@ func NewAciContainersOperator(
...
+			current_Obj := obj.(*corev1.ConfigMap)
+			if current_Obj.Name == "aci-operator-config" {
+				if err == nil {
+					config_map_queue.Add(key)
+				}
+			}
+
+		},
+		UpdateFunc: func(prevObj, currentObj interface{}) {
+			log.Info("In UpdateFunc for ConfigMap")
```
snaiksat, line 370: ditto

pkg/controller/acicontainersoperator.go
```diff
@@ -329,6 +353,35 @@ func NewAciContainersOperator(
...
+			current_Obj := currentObj.(*corev1.ConfigMap)
+			if current_Obj.Name == "aci-operator-config" {
+				controller.handleConfigMapUpdate(prevObj, currentObj, config_map_queue)
+			}
+		},
+		DeleteFunc: func(obj interface{}) {
+			key, err := cache.MetaNamespaceKeyFunc(obj)
+			log.Info("Deleted ConfigMap key is :", key)
```
snaiksat, line 378: ditto

pkg/controller/acicontainersoperator.go
```diff
@@ -379,10 +432,77 @@ func (c *Controller) handledaemonsetUpdate(oldobj interface{}, newobj interface{
...
+	}
+
+	log.Info("Reading current aci-operator-config configmap")
+	rawSpec, err := c.ReadConfigMap("/tmp/aci-operator.conf")
+	if err != nil {
+		return true
+	}
+
+	obj := c.CreateAciContainersOperatorObj()
+	log.Info("Unmarshalling the ConfigMaps...")
```
snaiksat, line 463: name of configmap...also notice how sometimes you use "Config-map" and then sometimes you say "ConfigMaps"...admittedly a nit, but there has to be consistency in conventions at least in the same module...

pkg/controller/acicontainersoperator.go
```diff
@@ -379,10 +432,77 @@ func (c *Controller) handledaemonsetUpdate(oldobj interface{}, newobj interface{
...
+		_, err = c.Operator_Clientset.AciV1alpha1().AciContainersOperators(os.Getenv("SYSTEM_NAMESPACE")).
+			Update(context.TODO(), acicnioperator, metav1.UpdateOptions{})
+		if err != nil {
+		}
+	}
+	return false
+}
+
+func (c *Controller) handleConfigMapUpdate(oldobj interface{}, newobj interface{}, queue workqueue.RateLimitingInterface) {
+	log.Info("In Config-Map update handler ")
```
snaiksat, line 487: name of configmap

pkg/controller/acicontainersoperator.go
```diff
@@ -401,6 +521,22 @@ func (c *Controller) ReadConfigMap(field string) ([]byte, error) {
...
+	if err != nil {
+		log.Error(err)
+		return err
+	}
+	err = ioutil.WriteFile(field, data_byte, 0777)
+	if err != nil {
+		log.Error(err)
+		return err
+	}
+	log.Info("Writing the Config-Map: ",data.Name," in the file")
```
snaiksat, line 536: There is no write happening after this log statement. Either the log statement should be "Wrote..." or we should move this log statement to before the write is done...

pkg/controller/acicontainersoperator.go
```diff
@@ -109,6 +112,9 @@ const aciContainersController = "aci-containers-controller"
 const aciContainersHostDaemonset = "aci-containers-host"
 const aciContainersOvsDaemonset = "aci-containers-openvswitch"
 
+var Aci_operator_config_path = "/usr/local/etc/aci-containers/aci-operator.conf"
```
snaiksat, line 115: while this addresses the earlier comment about predefining the strings, I am still not clear why this path has to be an absolute path whereas earlier it was not...

snaiksat (changes requested):

snaiksat:
And gofmt is still failing...

pkg/controller/acicontainersoperator.go
abhis2112: This can add an extra line of code to fetch the name of the config-map. It can be done if really needed. However in subsequent call in handler we mention which config-map

pkg/controller/acicontainersoperator.go
abhis2112: Earlier also it was. The function ReadConfigMap is called like below 

**rawSpec, err := c.ReadConfigMap("aci-containers/aci-operator.conf")**

and inside the function, it is coded with **/usr/local/etc**

**func (c *Controller) ReadConfigMap(field string) ([]byte, error) {
	raw, err := ioutil.ReadFile("/usr/local/etc/" + field)**

So eventually it becomes absolute and we are not changing this location as we mention this location in volume mount in deployment file

snaiksat (approved):

PR #919, opened 2021-10-14
Collect ipam details on various ports in cluster-report

vlella:


cmd/acikubectl/cmd/debug.go
```diff
@@ -515,12 +558,47 @@ func accVersionCmdArgs(systemNamespace string) []string {
...
+		"-c", "aci-containers-controller", "--", "curl", "-s", "localhost:8091/status"}
+}
+
 func aciContainerHostVersionCmdArgs(systemNamespace string) []string {
 	return []string{"-n", systemNamespace, "exec", "daemonsets/aci-containers-host",
 		"-c", "aci-containers-host", "-i", "--",
 		"aci-containers-host-agent", "--version"}
 }
 
+func hostIpamLogCmdArgs(systemNamespace string, podName string, containerName string, args []string) []string {
```
anmol372, line 577: Can we combine these 5 host functions, since it seems like all of them expect the same parameters.

anmol372, line 577: Can we combine the 5 host agent functions, since they are taking the same arguments and the way you have designed it should be able to handle modifications to the individual commands in future.

snaiksat (approved):

PR #924, opened 2021-11-02
Add checks for snat config map during update

tanyatukade:
- If there are no changes to snat cfgmap, don't clearSnatGlobalCache and don't add it to update queue, instead log and return


snaiksat:
Thanks for the work on this. Is it possible to add a UT for this (assuming the framework is already present)?

snaiksat (approved):

PR #928, opened 2021-11-18
Revert netpol revert

anmol372:


snaiksat:
May not be needed, closing for now, we can reopen if needed.

PR #929, opened 2021-12-06
Fix for snat policy enforcing wrong snat ip 

akhilamohanan:
List only the services which match the resource label instead of listing all the services in that namespace

Listing all services in the namespace and comparing its label with the labels of snatpolicies in cache was leading to updation of all snatpolicies in that namespace in the nodeinfo. Ideally it should update the nodeinfo only with snatpolicies corresponding to the resources in that node

### DIFF
--- a/pkg/hostagent/snats.go
+++ b/pkg/hostagent/snats.go
@@ -1011,11 +1011,7 @@ func (agent *HostAgent) getMatchingSnatPolicy(obj interface{}) (snatPolicyNames 
 				}
 				if res == POD {
 					if len(item.Spec.SnatIp) == 0 {
-						var services []*v1.Service
-						cache.ListAllByNamespace(agent.serviceInformer.GetIndexer(), namespace, labels.Everything(),
-							func(servobj interface{}) {
-								services = append(services, servobj.(*v1.Service))
-							})
+						services := util.GetServicesByOneOfLables(agent.serviceInformer.GetIndexer(), namespace, label)
 						// list the pods and apply the policy at service target
 						for _, service := range services {
 							if util.MatchLabels(item.Spec.Selector.Labels,
--- a/pkg/util/snat.go
+++ b/pkg/util/snat.go
@@ -22,8 +22,11 @@ import (
 	snatglobalclset "github.com/noironetworks/aci-containers/pkg/snatglobalinfo/clientset/versioned"
 	snatpolicy "github.com/noironetworks/aci-containers/pkg/snatpolicy/apis/aci.snat/v1"
 	snatpolicyclset "github.com/noironetworks/aci-containers/pkg/snatpolicy/clientset/versioned"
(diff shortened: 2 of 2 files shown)


pkg/hostagent/snats.go
```diff
@@ -1006,10 +1006,14 @@ func (agent *HostAgent) getMatchingSnatPolicy(obj interface{}) (snatPolicyNames
 				if res == POD {
 					if len(item.Spec.SnatIp) == 0 {
 						var services []*v1.Service
-						cache.ListAllByNamespace(agent.serviceInformer.GetIndexer(), namespace, labels.Everything(),
-							func(servobj interface{}) {
-								services = append(services, servobj.(*v1.Service))
-							})
+						for key, value := range label {
+							singleLabel := make(map[string]string)
```
tanyatukade, line 1010: the label in L976 is already a map[string][string], why do we need to create another map[string[string] with the same key, value pairs?

We can instead have selector := labels.SelectorFromSet(labels.Set(snatpolicy.Spec.Selector.Labels)) and use this selector while listing the services

akhilamohanan: @tanyatukade , label variable has multiple labels in it. we we select services using label variable, it will try to get the services which has all those labels in it. But, we need to include the services which has which matches with at least one of the pod labels. Thats why I'm getting labels one by one and adding the service that matches that label.

tanyatukade: We can have a similar func (https://github.com/noironetworks/aci-containers/blob/0791748e9d0eae3e7f67df67afa93d7c45bb74a0/pkg/controller/snats.go#L169)  in hostagent that will get all the serives using the selector. Can we try this?

tanyatukade: Since we have access to the pod labels, we can pass that as the selector argument

tanyatukade: Since SelectorFromSet selects all the labels from the map (i.e the entire set), we wont be able to use it to match individual labels. We should check if the library has any other funcs that selects individual labels instead of the entire set, that way we can avoid the for loop.

akhilamohanan: @tanyatukade , I tried multiple library functions, but couldn't find any function that gets/lists the services that matches one of the given labels. All of them are listing the services that has all the given labels.

tanyatukade: Thanks for checking, can we add a function that can do this in our utils package? And call that func here?

akhilamohanan: Addressed

tanyatukade (approved):
LGTM

akhilamohanan:
https://github.com/noironetworks/aci-containers/pull/990 -- handled here

akhilamohanan:
https://github.com/noironetworks/aci-containers/pull/990 -- handled here

PR #939, opened 2021-12-10
Split subscription DNs

shastrinator:
split subscription DNs when cardinality is marked high for any child
tagAnnotations cannot be split.
Each child Dn will be subscribed to as a separate Dn in this model.


snaiksat:
Thanks @shastrinator . Keeping in mind that the there is a limit on the count of objects returned by the APIC, is there a value in retrieving the count as well from the queries and then logging a warning message when the count goes above a certain threshold, say 10k?

snaiksat (approved):

PR #940, opened 2021-12-16
Enhanced the logs in network attachment definition code base

Bhavanaashok33:
Signed-off-by: Bhavanaashok33 <bhavana.babu32@gmail.com>

### DIFF
--- a/pkg/hostagent/netattachdef.go
+++ b/pkg/hostagent/netattachdef.go
@@ -19,7 +19,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"os"
-	"path/filepath"
+	"path"
 	"strings"
 	"time"
 
@@ -131,7 +131,7 @@ func (agent *HostAgent) networkAttDefChanged(ntd *netpolicy.NetworkAttachmentDef
 					}
 				}
 			} else {
-				agent.log.Debug("network atttachment does not specify opflex-agent-cni and opflex-agent-cni-ipam")
+				agent.log.Debug("Network Attachment Definition is not ACICNI specific")
 			}
 		}
 	}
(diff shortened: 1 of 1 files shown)


pkg/hostagent/netattachdef.go
```diff
@@ -175,11 +175,11 @@ func (agent *HostAgent) getPodResource(metadata *md.ContainerMetadata) error {
 		if agent.netattdefmap[netAttName] != nil {
 			isAcicniNetwork = true
 		} else {
-			return nil
+			return fmt.Errorf("Network Attachment Definition CR not applied: Must mention ACICNI plugin")
 		}
 	}
 	if isAcicniNetwork {
-		podResourceSock := filepath.Join(kubeletPodResourceDefaultPath, podresources.Socket+".sock")
+		podResourceSock := path.Join(kubeletPodResourceDefaultPath, podresources.Socket+".sock")
```
mchalla, line 182: shouldnt we be using filepath when working with files ?

mchalla (commented):
Looks good to me. see one comment on usage of path.

pkg/hostagent/netattachdef.go
Bhavanaashok33: For Linux Operating systems, either of FilePath or Path package can be used. Paths are always separated by forward slashes, such as the paths in URLs in Linux. 
The Filepath package uses either forward slashes or backslashes, depending on the operating system. To process paths such as URLs that always use forward slashes regardless of the operating system, we use path package.
Usage of filepath or path doesnt make any difference for linux operation  :-)  I can revert it filepath anyway.

mchalla: I was just curious. either is fine to me.

mchalla (approved):

PR #941, opened 2021-12-21
Fix for IPAM leaks in hostagent due to data races

akhilamohanan:
Added locks to fix data races in hostagent.

Accessing shared variables without lock in hostagent was causing data
races leading to IPAM leaks


tanyatukade:
@akhilamohanan  Has this been tested against the race conditions detected?

pkg/hostagent/status.go
```diff
@@ -68,6 +68,7 @@ func (agent *HostAgent) RunStatus() {
 	http.HandleFunc("/status", func(w http.ResponseWriter, r *http.Request) {
 		w.Header().Set("Content-Type", "application/json")
 		agent.indexMutex.Lock()
+		agent.ipamMutex.Lock()
```
snaiksat, line 71: What is the relevance of adding a second mutex here? You already have one mutex which you are locking and unlocking at the same place...

pkg/hostagent/status.go
```diff
@@ -78,11 +79,13 @@ func (agent *HostAgent) RunStatus() {
 			UsedIPs: len(agent.usedIPs),
 		}
 		json.NewEncoder(w).Encode(status)
+		agent.ipamMutex.Unlock()
 		agent.indexMutex.Unlock()
 	})
 	http.HandleFunc("/ipam", func(w http.ResponseWriter, r *http.Request) {
 		w.Header().Set("Content-Type", "application/json")
 		agent.indexMutex.Lock()
+		agent.ipamMutex.Lock()
```
snaiksat, line 88: What is the relevance of adding a second mutex here? You already have one mutex which you are locking and unlocking at the same place...

akhilamohanan:
> @akhilamohanan Has this been tested against the race conditions detected?

Yes @tanyatukade . Verified with the same script that was used to reproduce the issue.

pkg/hostagent/status.go
akhilamohanan: indexMutex is acquired when we access cache in hostagent pkg (agent.usedIP in this case).
ipamMutex is acquired when we access shared variable in ipam pkg.(agent.podIps.CombineV4())

tanyatukade: I dont think we would need both locks, it should be only one of them. Since it is same type of lock (sync.Mutex).

akhilamohanan:
Hostagent is built with race flag and pushed with tag  ipam-leak-race (quay.io/noirolabs/aci-containers-host:ipam-leak-race)

Tests done:
1) pytests
2) /home/noiro/v1_version_test/test_script.sh script on fab14

snaiksat (approved):

tanyatukade (approved):

snaiksat (approved):

PR #945, opened 2022-01-17

akhilamohanan:
Added indexMutex lock while updating snat-uuid in epfile

### DIFF
--- a/pkg/hostagent/snats.go
+++ b/pkg/hostagent/snats.go
@@ -223,6 +223,8 @@ func (agent *HostAgent) initSnatPolicyInformerBase(listWatch *cache.ListWatch) {
 }
 
 func (agent *HostAgent) snatPolicyAdded(obj interface{}) {
+	agent.indexMutex.Lock()
+	defer agent.indexMutex.Unlock()
 	agent.snatPolicyCacheMutex.Lock()
 	defer agent.snatPolicyCacheMutex.Unlock()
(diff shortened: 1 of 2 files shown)


akhilamohanan:
Testing is in progress

akhilamohanan:
Tests done:
pytests 
stress test (by deleting host-agent in an interval of 5 min)

Built hostagent images with the patch:
quay.io/noirolabs/aci-containers-host:snat-uuid  (without enabling race flag)
quay.io/noirolabs/aci-containers-host:snat-uuid-race (by enabling race flag)

pkg/hostagent/snats.go
```diff
@@ -1200,15 +1206,20 @@ func (agent *HostAgent) isPolicyNameSpaceMatches(policyName string, namespace st
...
-	agent.indexMutex.Unlock()
 	if check {
-		return val.PlcyUuids
-
-	} else {
-		return []string{}
+		err := util.DeepCopyObj(val, localInfo)
+		if err != nil {
+			agent.log.Error(err.Error())
+			return plcyUuids
```
tanyatukade, line 1218: Why are we returning the plcyUuids even if the deep copy fails? We should be returning nil. In general it is best practice to have the return value([]string) and the error in the func definition, so we can return both. For this case, when there is an err we return nil, err and when the func runs sucessfully we return plcyUuids, nil

rupeshsahuoc: addressed

PR #946, opened 2022-01-21
Fixed deadlock

akhilamohanan:
Fixed order of unlock in UpdateLocalInfoCr()

### DIFF
--- a/pkg/hostagent/snatlocalinfo.go
+++ b/pkg/hostagent/snatlocalinfo.go
@@ -51,7 +51,6 @@ func (agent *HostAgent) UpdateLocalInfoCr() bool {
 		var localInfo SnatLocalInfo
 		localInfo.snatIp = ginfo.SnatIp
 		agent.snatPolicyCacheMutex.RLock()
-		defer agent.snatPolicyCacheMutex.RUnlock()
 		if _, ok := agent.snatPolicyCache[ginfo.SnatPolicyName]; ok {
 			if len(agent.snatPolicyCache[ginfo.SnatPolicyName].Spec.DestIp) == 0 {
 				localInfo.destIps = []string{"0.0.0.0/0"}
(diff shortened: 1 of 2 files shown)


pkg/hostagent/snatlocalinfo.go
```diff
@@ -60,6 +59,7 @@ func (agent *HostAgent) UpdateLocalInfoCr() bool {
 					agent.snatPolicyCache[ginfo.SnatPolicyName].Spec.DestIp
 			}
 		}
+		agent.snatPolicyCacheMutex.RUnlock()
```
snaiksat, line 62: Is this the right place to move this unlock? We are iterating over agent.opflexSnatLocalInfos in L67. What happens if we leave the defer as is, or we unlock right before unlock the indexMutex in L93?

akhilamohanan: As far as I know, snatPolicyCacheMutex is acquired while accessing snatPolicyCache. 

Leaving defer as is will lead to deadlock as the lock is acquired in loop. After the 1st iteration(ie, after first rlock is acquired), if some other go routine starts waiting for write lock, then the 2nd iteration will cause deadlock as it is trying to access 2nd rlock without releasing 1st.

unlock right before unlock the indexMutex in L93 -- same reason as above and since the rlock is accessed in a loop, a single unlock is not sufficient.

Another thing we can do is, access rlock just before for loop(L50) and unlock it just above index unlock(L93).

snaiksat: I missed the point that this is in a loop (should have expanded the prior code). In that case I am fine leaving it the way you have it. However, which lock is protecting agent.opflexSnatLocalInfos, indexMutex?

akhilamohanan: Yes. indexMutex protects agent.opflexSnatLocalInfos. [(For ref)](https://github.com/noironetworks/aci-containers/blob/master/pkg/hostagent/snats.go#L1204)

snaiksat (approved):

PR #952, opened 2022-02-02
excess log fix in aci controller

Vikash082:
Added condition to log device details only when the state of device gets transitioned to "connected" state.


snaiksat (approved):

snaiksat:
thanks @Vikash082, can we also backport this to kmr2 branch?

PR #967, opened 2022-03-03

akhilamohanan:
* RCA: Sometimes syncEps() was called at the time of initialization even before
opflexSnatLocalinfo was not updated with snat-uuids and empty snat-uuid list was written to ep file.

* Fix: If opflexSnatLocalinfo update is not happened before syncEps() at the
time of initialization, snat-uuids in the ep file will be maintained

### DIFF
--- a/pkg/hostagent/environment.go
+++ b/pkg/hostagent/environment.go
@@ -177,29 +177,48 @@ func (env *K8sEnvironment) PrepareRun(stopCh <-chan struct{}) (bool, error) {
 	cache.WaitForCacheSync(stopCh, env.agent.nodeInformer.HasSynced)
 	env.agent.log.Info("Node cache sync successful")
 
+	env.agent.log.Debug("Starting service informer")
+	go env.agent.serviceInformer.Run(stopCh)
+	env.agent.log.Info("Waiting for service cache sync")
+	cache.WaitForCacheSync(stopCh, env.agent.serviceInformer.HasSynced)
+	env.agent.log.Info("Service cache sync successful")
+
+	env.agent.log.Debug("Starting snat global informer")
+	go env.agent.snatGlobalInformer.Run(stopCh)
+	env.agent.log.Info("Waiting for snat global cache sync")
+	cache.WaitForCacheSync(stopCh, env.agent.snatGlobalInformer.HasSynced)
+	env.agent.log.Info("Snat global cache sync successful")
+
+	env.agent.log.Debug("Starting snat policy informer")
+	go env.agent.snatPolicyInformer.Run(stopCh)
+	env.agent.log.Info("Waiting for snat policy sync")
+	cache.WaitForCacheSync(stopCh, env.agent.snatPolicyInformer.HasSynced)
+	env.agent.log.Info("Snat policy sync successful")
+
+	env.agent.log.Debug("Starting rdConfig informer")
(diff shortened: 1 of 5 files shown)


pkg/hostagent/environment.go
```diff
@@ -177,29 +177,46 @@ func (env *K8sEnvironment) PrepareRun(stopCh <-chan struct{}) (bool, error) {
...
-	go env.agent.serviceInformer.Run(stopCh)
 	go env.agent.nsInformer.Run(stopCh)
 	go env.agent.netPolInformer.Run(stopCh)
 	go env.agent.depInformer.Run(stopCh)
 	go env.agent.rcInformer.Run(stopCh)
-	go env.agent.snatGlobalInformer.Run(stopCh)
-	go env.agent.snatPolicyInformer.Run(stopCh)
 	go env.agent.qosPolicyInformer.Run(stopCh)
-	go env.agent.rdConfigInformer.Run(stopCh)
-	go env.agent.netAttDefInformer.Run(stopCh)
```
snaiksat, line 195: We seem to be missing this in the refactored code...

pkg/hostagent/snatlocalinfo.go
```diff
@@ -32,6 +32,30 @@ type SnatLocalInfo struct {
...
+	snatLocalInfoClient := env.snatLocalInfoClient
+	snatLocalInfoCr, err := snatLocalInfoClient.AciV1().SnatLocalInfos(agent.config.AciSnatNamespace).Get(context.TODO(), agent.config.NodeName, metav1.GetOptions{})
+	if err != nil {
+		agent.log.Error("Failed to get snatlocalinfo ", err.Error())
+		return err
+	}
+	for _, localInfo := range snatLocalInfoCr.Spec.LocalInfos {
+		if localInfo.PodUid != "" {
+			agent.indexMutex.Lock()
+			_, ok := agent.opflexSnatLocalInfos[localInfo.PodUid]
```
snaiksat, line 46: nit: I don't think step is necessary, but don't need to change if everything else is working

pkg/hostagent/pods.go
```diff
@@ -472,16 +472,16 @@ func (agent *HostAgent) syncEps() bool {
 			if seen[ep.Uuid] {
 				continue
 			}
+			epfile := agent.FormEPFilePath(ep.Uuid)
 			poduuid := strings.Split(ep.Uuid, "_")[0]
-			ep.SnatUuid, err = agent.getSnatUuids(poduuid)
+			ep.SnatUuid, err = agent.getSnatUuids(poduuid, epfile)
```
snaiksat, line 477: Isn't this the case where the EP file does not exist and we are creating a new one? If so, I guess we should not return error in getSnatUuids() if the file is not present...

snaiksat (changes requested):

pkg/hostagent/environment.go
akhilamohanan: addressed

pkg/hostagent/snatlocalinfo.go
akhilamohanan: addressed

pkg/hostagent/pods.go
akhilamohanan: addressed

snaiksat (approved):

PR #970, opened 2022-03-03

akhilamohanan:
* RCA: Sometimes syncEps() was called at the time of initialization even before
opflexSnatLocalinfo was not updated with snat-uuids and empty snat-uuid list was written to ep file.

* Fix: If opflexSnatLocalinfo update is not happened before syncEps() at the
time of initialization, snat-uuids in the ep file will be maintained

(cherry picked from commit c15e2e5dba474e20c430575fdd651f91b9c02652)

### DIFF
--- a/pkg/hostagent/snatlocalinfo.go
+++ b/pkg/hostagent/snatlocalinfo.go
@@ -18,6 +18,7 @@ package hostagent
 
 import (
 	"context"
+	"errors"
 	"reflect"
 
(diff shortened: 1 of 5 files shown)


pkg/hostagent/snatlocalinfo.go
```diff
@@ -32,6 +33,31 @@ type SnatLocalInfo struct {
 	snatpolicyName string
 }
 
+func (agent *HostAgent) populateSnatLocalInfos() error {
+	env := agent.env.(*K8sEnvironment)
```
snaiksat, line 37: Perhaps we need some log statement(s) in this function to show how this function was exercised in the working case.

PR #986, opened 2022-03-15
Fix for missing snat-uuid in ep file

akhilamohanan:
Fix for missing snat-uuid in ep-file when a pod of deployment and
hostagent restarts at same time

A case where deloyment doesnt have snat association but a service
corresponding to the deployment have snat association was not handled
and hence the issue.

(cherry picked from commit a54b39769c85c2ed31556b3ed6c85cfda2282421)


akhilamohanan:
Testing in progress

snaiksat (approved):

PR #1007, opened 2022-04-28
Single request for hpp

akhilamohanan:


pkg/controller/network_policy.go
```diff
@@ -1060,6 +1081,99 @@ func (cont *AciController) buildServiceAugment(subj apicapi.ApicObject,
 	}
 }
 
+func (cont *AciController) handleMulNetPolUpdate(nps []interface{}) bool {
```
shastrinator, line 1084: Can we take a bool argument to call WriteMulApicObjects and use the existing handleNetPolUpdate function?

pkg/apicapi/apic_sync.go
```diff
@@ -137,6 +137,113 @@ func (conn *ApicConnection) apicObjCmp(current ApicObject,
...
+		return
+	}
+	for _, val := range *deleteBody {
+		dn, okdn := val.Attributes["dn"]
+		if okdn && dn == deldn {
+			attrs := make(map[string]interface{})
+			attrs["dn"] = val.Attributes["dn"]
+			attrs["status"] = "deleted"
+			val.Attributes = attrs
+			conn.log.Debug("deleteBody: ", *deleteBody)
```
shastrinator, line 151: Assuming all the logs will be cleaned up with another pass

pkg/apicapi/apic_metadata.go
```diff
@@ -154,8 +154,7 @@ func injectedSvcPortNormalizer(b *ApicObjectBody) {
 var metadata = map[string]*apicMeta{
 	"fvTenant": {
 		attributes: map[string]interface{}{
-			"name":      "",
-			"nameAlias": "",
+			"name": "",
```
shastrinator, line 157: Was this change necessary?

pkg/controller/network_policy.go
akhilamohanan: Existing handleNetPolUpdate is handling one np at a time. Here we need multiple nps at a time

pkg/apicapi/apic_sync.go
akhilamohanan: Yes. I will remove all the debug logs that is not necessary.

pkg/apicapi/apic_metadata.go
akhilamohanan: Yes. Since we are sending POST request to <tenant>.json and in the request body, if nameAlias is empty it was replacing the tenant alias (openupi) in our case to empty string

pkg/controller/network_policy.go
shastrinator: Each inner loop is the handleNetPolUpdate function. Can we not call in a loop and omit the writing to Apic part with a flag?

akhilamohanan: Moved common code to a separate function.

shastrinator (approved):
LGTM

akhilamohanan:
Handled in https://github.com/noironetworks/aci-containers/pull/1015

PR #1016, opened 2022-05-25
Add ep to service graph only after ep is READY

akhilamohanan:
Service graph was programmed when endpoint was updated. We were not
making sure that ep is in ready state.

Added check to program service graph only when ep is ready

### DIFF
--- a/pkg/controller/services.go
+++ b/pkg/controller/services.go
@@ -1639,21 +1639,30 @@ func (cont *AciController) endpointSliceUpdated(oldobj interface{}, newobj inter
 		cont.log.Error("error processing Endpointslice object: ", newobj)
 		return
 	}
+	proceed := true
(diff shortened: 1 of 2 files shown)


pkg/controller/services.go
```diff
@@ -1639,21 +1639,30 @@ func (cont *AciController) endpointSliceUpdated(oldobj interface{}, newobj inter
...
 		return
 	}
+	proceed := true
+	for _, endpoint := range newendpointslice.Endpoints {
+		if endpoint.Conditions.Ready == nil || !*endpoint.Conditions.Ready {
+			proceed = false
+		}
+	}
+	if !proceed {
+		cont.log.Error("New enpoints are not in ready state")
```
snaiksat, line 1649: should this be an Error log? wouldn't we hit this point even when the pod is in Pending state?

snaiksat (approved):

PR #1027, opened 2022-06-30
Installing and configuring lsyncd service to host-agent pod and adding dpu support to host-agent

Bhavanaashok33:
Purpose of Lsyncd service is that it is capable of syncing opflex agent filesystems between dpu and the host and This is because we are running opflex-agent on the DPU and host-agent pod on the host(Basically on two difference machine unlike before).

### DIFF
--- a/pkg/hostagent/config.go
+++ b/pkg/hostagent/config.go
@@ -236,6 +236,9 @@ type HostAgentConfig struct {
 
 	// Sriov and Ovs Hardward Offload enabled
 	OvsHardwareOffload bool `json:"enable-ovs-hw-offload,omitempty"`
+
+	// DpuOvsDBSocket when OpflexMode is dpu selects ovsdb sock on dpu
+	DpuOvsDBSocket string `json:"dpu-ovsdb-socket,omitempty"`
 }
 
 func (config *HostAgentConfig) InitFlags() {
@@ -307,4 +310,5 @@ func (config *HostAgentConfig) InitFlags() {
 	flag.BoolVar(&config.EnableNodePodIF, "enable-nodepodif", false, "Enable NodePodIF")
 	flag.StringVar(&config.EPRegistry, "ep-registry", "", "Enable PodIF")
 	flag.BoolVar(&config.OvsHardwareOffload, "enable-sriov-config", false, "SRIOV config and ovs hardware offload feature")
+	flag.StringVar(&config.DpuOvsDBSocket, "dpu-ovsdb-socket", "tcp:192.168.200.2:6640", "TCP socket on DPU to connect to")
 }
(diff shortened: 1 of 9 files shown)


pkg/hostagent/config.go
```diff
@@ -307,4 +310,5 @@ func (config *HostAgentConfig) InitFlags() {
 	flag.BoolVar(&config.EnableNodePodIF, "enable-nodepodif", false, "Enable NodePodIF")
 	flag.StringVar(&config.EPRegistry, "ep-registry", "", "Enable PodIF")
 	flag.BoolVar(&config.OvsHardwareOffload, "enable-sriov-config", false, "SRIOV config and ovs hardware offload feature")
+	flag.StringVar(&config.DpuOvsDBSocket, "dpu-ovsdb-socket", "tcp:192.168.200.2:6640", "TCP socket on DPU to connect to")
```
snaiksat, line 313: Is it okay to default to this specific address? How did we choose this one?

pkg/hostagent/opflex.go
```diff
@@ -86,6 +86,17 @@ func (agent *HostAgent) discoverHostConfig() (conf *HostAgentNodeConfig) {
 		conf.OpflexPeerIp = "127.0.0.1"
 		agent.log.Debug("\n  == Opflex: Running in overlay mode ==\n")
 		return
+	} else if agent.config.OpflexMode == "dpu" {
+		conf = &HostAgentNodeConfig{}
+		conf.VxlanIface = "bond0.4093"
+		conf.UplinkIface = "bond0"
+		conf.VxlanAnycastIp = "10.0.0.32"
+		conf.OpflexPeerIp = "10.0.0.30"
```
snaiksat, line 94: Why are the above values hardcoded, would they not be system dependent?

pkg/hostagent/config.go
mchalla: Sumit, those are part of my changes. yeah it should be ok because this is the private network between DPU and host and it needs to be setup before the host can talk to the DPU and since the customer will not have access to the DPU there is no way for him to set this up via any config we generate. This code review is missing changes on DPU itself so its not clear without that context.

pkg/hostagent/opflex.go
mchalla: That is hardcoded because the bond interface is not present on the host. Its present on the DPU and again a customer cannot provision this currently because he has no access to DPU. I thought these do not change since I have never seen a different value for them in any config. But I might be wrong. Again the DPU side has its own config thats missing in this review. One way to make these dynamic would be to get them from the k8s setup we do on the DPU. There are 2 k8s clusters running, one on DPU that we control and another on host. Thanks.

mchalla: But its a more involved change because it involves.
1. DPU using whatever from the k8s config it received.
2. Host agent sending a remote ip link command to dpu and populating those values.
3
Technically I do not see why the host agent should even care what those values are besides the sanity check. But today if that isnt populated host agent will fail to come up. So if thats the case we can do the sanity check on DPU and skip 2 that is the more involved change.

snaiksat: Thanks Madhu. Assuming that the 4093 here corresponds to the infra VLAN, that could definitely vary from one setup to another. Anyway, I will merge these changes for now with the understanding that a follow up PR for DPU changes is on the way and can provide any changes if required.

snaiksat (approved):

PR #1033, opened 2022-08-03
Do not wait for service ep to be ready for conformance tests

akhilamohanan:
Made waiting for serice ep to be ready before adding it to the service
graph configurable. Default behaviour is to wait for ep to be ready.
For conformance tests, it will not wait.

### DIFF
--- a/pkg/controller/services.go
+++ b/pkg/controller/services.go
@@ -1663,15 +1663,17 @@ func (cont *AciController) endpointSliceUpdated(oldobj interface{}, newobj inter
 		cont.log.Error("error processing Endpointslice object: ", newobj)
 		return
 	}
-	proceed := true
-	for _, endpoint := range newendpointslice.Endpoints {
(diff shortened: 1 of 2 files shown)


pkg/controller/services.go
```diff
@@ -1663,15 +1664,17 @@ func (cont *AciController) endpointSliceUpdated(oldobj interface{}, newobj inter
 		cont.log.Error("error processing Endpointslice object: ", newobj)
 		return
 	}
-	proceed := true
-	for _, endpoint := range newendpointslice.Endpoints {
-		if endpoint.Conditions.Ready == nil || !*endpoint.Conditions.Ready {
-			proceed = false
+	if strings.ToLower(os.Getenv("CONFORMANCE_TESTS")) != "true" {
```
snaiksat, line 1667: Hate to nit-pick here, can we call this "wait_for_service_ep_readiness"? Also can we drive this change from config.go with the default value set to True? That way we can hook up with acc-provision.

snaiksat (approved):

PR #1043, opened 2022-09-07
Add USER 1001 at end of dockerfiles to support Container certificatio…

vlella:
…n tooling for Redhat operartor certification workflow


vlella:
this PR not required anymore after convo with RH on slack channel.

PR #1045, opened 2022-09-15
Upgraded deprecated packages

akhilamohanan:
Upgraded deprecated packages and the packages that were using the deprecated packages.
Compilation needs go 1.19


akhilamohanan:
Testing pending

snaiksat (approved):

PR #1054, opened 2022-10-26
To avoid truncating of information, use "get" instead of "describe"

akhilamohanan:
(cherry picked from commit 29b92f582f74ee8850af9afa4c9216db09cbf09f)


vlella:
thanks @akhilamohanan for pointing this missing backport to keep in sync with acc-pytest kmr2 branch.

PR #1068, opened 2023-02-09
update prometheus-cpp to v1.0.1 in docker files

gautvenk:
Signed-off-by: Gautam Venkataramanan <gautam.chennai@gmail.com>


tomflynn (approved):

snaiksat (approved):

snaiksat:
Thanks @gautvenk , our stable branch currently is kmr2, would it make sense to backport this? If it addresses CVE and is a backward compatible change, I would suggest that we backport. (Same for the opflex side change.)

gautvenk:
> Thanks @gautvenk. Our stable branch currently is kmr2; would it make sense to backport this? If it addresses CVE and is a backward compatible change, I would suggest that we backport. (Same for the opflex side change.)

Sounds good @snaiksat, i will port the changes to kmr2.

PR #1075, opened 2023-03-17
Custom sg

akhilamohanan:


pkg/controller/services.go
```diff
@@ -684,6 +711,16 @@ func (cont *AciController) updateServiceDeviceInstance(key string,
 	}
 
 	graphName := cont.aciNameForKey("svc", "global")
+	deviceName := cont.aciNameForKey("svc", "global")
```
shastrinator, line 714: How would you pass a different name here? With another annotation?

pkg/apicapi/apic_sync.go
```diff
@@ -38,6 +38,15 @@ func (conn *ApicConnection) apicBodyAttrCmp(class string,
 		bodyd.Attributes = make(map[string]interface{})
 	}
 	for p, def := range meta.attributes {
+		if class == "vzRsSubjGraphAtt" && p == "tnVnsAbsGraphName" {
+			_, forceResolveExists := bodyc.Attributes["forceResolve"]
```
shastrinator, line 42: As I understand, this allows the user to change the graph associated with the contract. So the user has to necessarily create a new service graph with the FW inserted? Should we not have an exclusion case for vnsAbsGraph as well..

snaiksat: > So the user has to necessarily create a new service graph with the FW inserted?

Yes.

pkg/controller/services.go
akhilamohanan: If user annotates the service, we will check the tnVnsAbsGraphName in the contract (https://github.com/noironetworks/aci-containers/pull/1075/files#diff-d5c31f7209190065692316305836fc8a42834003000db2714b32ac867cc5c96fR717) and use the same graphName that is there in apic

pkg/apicapi/apic_sync.go
akhilamohanan: vnsAbsGraph is created by user (Not editing the one created by cni ), so there will not be any aci cni annotations and we will not process it. But User edits the graphName of the contract created by aci cni, and since it has annotations, we we process it and we have to ignore the field while processing

pkg/controller/services.go
shastrinator: This question was why we need two arguments into ApicDevCtx, if we only use devicename argument.

akhilamohanan: apicDevCtx has reference of both deviceName and graphName. 
https://github.com/noironetworks/aci-containers/pull/1075/files#diff-d5c31f7209190065692316305836fc8a42834003000db2714b32ac867cc5c96fR576 - Here we use graphName
https://github.com/noironetworks/aci-containers/pull/1075/files#diff-d5c31f7209190065692316305836fc8a42834003000db2714b32ac867cc5c96fR578 - Here we use deviceName 

When we create a global SG template and device, we create it with same name <aci_prefix>_global_svc. So, we were passing just graphName (since graphName was same as deviceName) to the function apicDevCtx. But, in case of custom SG, device remains same(<aci_prefix>_global_svc) and graphName will be different. So, we had to pass both deviceName and graphName to apicDevCtx

shastrinator (approved):
LGTM

akhilamohanan:
Handled the changes in below PRs:
https://github.com/noironetworks/aci-containers/pull/1077
https://github.com/noironetworks/aci-containers/pull/1076

Closing the PR as it is directly on kmr2

PR #1076, opened 2023-03-21
Permits user to provide their own Service Graph Template in contract

akhilamohanan:
If user wants to use their own service graph template for an lb service, they need to annotate the service :
opflex.cisco.com/service-graph-name: <some-value>
and then modify the SG template reference in contract specific to the service to have their own SG template. In that case, controller will not revert back the reference to the default one that is created by cni


snaiksat:
Since #1075 was approved by @shastrinator , merging this.

snaiksat (approved):

PR #1077, opened 2023-03-21
Permits user to provide their own Service Graph Template in contract

akhilamohanan:
If user wants to use their own service graph template for an lb service, they need to annotate the service :
opflex.cisco.com/service-graph-name: <some-value>
and then modify the SG template reference in contract specific to the service to have their own SG template. In that case, controller will not revert back the reference to the default one that is created by cni

(cherry picked from commit ccfb80795a34f36ab8a25f3217617f92d9972c23)


snaiksat:
Let's hold off on merging this one since this is a stable branch.

snaiksat (approved):

PR #1080, opened 2023-04-25
Multipod

akhilamohanan:


pkg/controller/services.go
```diff
@@ -300,13 +300,139 @@ func deleteDevicesFromList(delDevices apicapi.ApicSlice, devices apicapi.ApicSli
...
+		cont.log.Debug("Failed to get APIC response, err: ", err.Error())
+		return "", err
+	}
+	for _, obj := range apicresp.Imdata {
+		for _, body := range obj {
+			for _, child := range body.Children {
+				for class, cbody := range child {
+					if class == "fvRsCEpToPathEp" {
+						tDn, ok := cbody.Attributes["tDn"].(string)
+						if ok {
```
shastrinator, line 334: It is possible that another cluster can have an overlapping ip. Please check here that either Dn of fvRsCEpToPathEp has the tenant name as uni/<tn-name>/xxx or the parent fvCEp has it in its Dn.

akhilamohanan:
https://github.com/noironetworks/aci-containers/pull/1084 -- Changes handled here.
Closing the PR as it is directly on kmr2

PR #1084, opened 2023-05-02
Support multipod vm migration

akhilamohanan:
Controller reports hostagent regarding the new aci pod the vm is migrated to and the hostagent will do dhcp release and renew so that the vlan interface will get new ip from the new pod subnet


akhilamohanan:
Please note that host agent changes will work only after building new host agent base image with these changes

snaiksat (approved):

PR #1100, opened 2023-06-14
Remove unused metrics and prom dep

tomflynn:
These were only referenced by test code


gautvenk (approved):
lgtm. Not sure why this was added in the first place:)

PR #1104, opened 2023-06-16
Fixes to support multipod vm migration

akhilamohanan:
* added delay between dhcp release and renew
* multipod specific code will be executed only when aci-multipod is set


pkg/apicapi/apic_sync.go
```diff
@@ -308,6 +308,10 @@ func (conn *ApicConnection) fullSync() {
 		"updates": len(updates),
 		"deletes": len(deletes),
 	}).Info("APIC full sync completed")
+
+	conn.SyncMutex.Lock()
+	conn.SyncDone = true
+	conn.SyncMutex.Unlock()
```
snaiksat, line 314: I am missing the point of this lock.

snaiksat (approved):

pkg/apicapi/apic_sync.go
akhilamohanan: https://github.com/noironetworks/aci-containers/pull/1104/files#diff-d5c31f7209190065692316305836fc8a42834003000db2714b32ac867cc5c96fR373 - SyncDone is used here also and this is a different thread.

snaiksat: What I meant to ask is that this seems to be an atomic action, so why is a lock required?

akhilamohanan: Even if it is atomic, if we are writing to SyncDone without accessing lock, there is a chance of race if this shared variable is read in the other thread concurrently

PR #1111, opened 2023-06-22
Fix check for deleted endpoint slice

tomflynn:
I'm not sure if there's a functional issue here but the linter pointed out the existing check is useless

@akhilamohanan please take a look. Another option would to just remove the check if it's not needed and the case is handled elsewhere


akhilamohanan (approved):

akhilamohanan:
Thanks @tomflynn . It should be if len(newendpointslice.Endpoints) < len(oldendpointslice.Endpoints)

snaiksat:
Good job linter (and Tom), bad job reviewer (note to self) ;-)

PR #1121, opened 2023-07-10
Add initial linter config

tomflynn:
Address many of the linter complaints
Disable some of the noisier linter categories

Linter still needs to be run manually as there are a number of issues still outstanding
Hopefully at some point in the future it can be run with every build


snaiksat:
This is great, thanks adding the linter and also for fixing some of the prominent issues.

snaiksat (approved):

PR #1142, opened 2023-07-31
Added opflex-device-reconnect-wait-timeout parameter

akhilamohanan:
User can configure opflex-device-reconnect-wait-timeout - the timeout in seconds to wait for reconnect when opflexOdev is diconnected for a node before triggering a dhcp release and renew of vlan interface


akhilamohanan:
Testing pending

snaiksat (approved):

PR #1145, opened 2023-08-02
Build chained image on travis

shastrinator:
All the changes are already in place.
Only modifying makefile/Dockerfile.


docker/travis/Dockerfile-host
```diff
@@ -11,6 +11,7 @@ RUN yum install --disablerepo=* --enablerepo=ubi-9-appstream-rpms --enablerepo=u
 RUN yum update --disablerepo=* --repofrompath=centos,https://mirror.stream.centos.org/9-stream/BaseOS/x86_64/os --repofrompath=centos-app,https://mirror.stream.centos.org/9-stream/AppStream/x86_64/os -y --nogpgcheck && rm -rf /var/cache/yum
 RUN yum install --disablerepo=* --repofrompath=centos,https://mirror.stream.centos.org/9-stream/BaseOS/x86_64/os --repofrompath=centos-app,https://mirror.stream.centos.org/9-stream/AppStream/x86_64/os dhcp-client iptables-nft -y --nogpgcheck && rm -rf /var/cache/yum
 
+RUN yum install --disablerepo=* --repofrompath=centos,https://mirror.stream.centos.org/9-stream/BaseOS/x86_64/os --repofrompath=centos-app,https://mirror.stream.centos.org/9-stream/AppStream/x86_64/os jq nmstate -y --nogpgcheck && rm -rf /var/cache/yum
```
snaiksat, line 14: is there a reason why we could not have added this in the preceding yum install?

shastrinator: just wanted this line to fail explicitly if it did. Now that it succeeds, I can append it to the earlier list.

snaiksat: Only the UT part of the travis job has been exercised; the dockerfile gets exercised only when a tag is added. I will proactively merge this.

snaiksat (approved):

PR #1163, opened 2023-09-08
Fix CVEs and update packages

abhis2112:


abhis2112:
Provided in a separate PR

PR #1182, opened 2023-10-18
add option to set logging level to debug for aci container operator

PariyaAshok:


pkg/controller/acicontainersoperator.go
```diff
@@ -54,6 +54,20 @@ import (
...
 )
 
+func init() {
+	lvl, ok := os.LookupEnv("OPERATOR_LOGGING_LEVEL")
+	// LOG_LEVEL not set, let's default to info
+	if !ok {
+		lvl = "info"
+	}
+	ll, err := log.ParseLevel(lvl)
+	if err != nil {
```
abhis2112, line 64: do we expect errors even after having debug or info level set in variable lvl. Or is it just done for in case of any error. I don't expect any error to be present after the above if check

abhis2112: we can optimize the code a little bit. like this:
lvl, ok := os.LookupEnv("OPERATOR_LOGGING_LEVEL")
	// LOG_LEVEL not set, let's default to info
	if !ok {
		log.SetLevel(log.InfoLevel)
	}else{
                 log.SetLevel(log.DebugLevel)   
        }

PariyaAshok: Thanks Abhishek, We need this check to catch invalid logrus Level provided by user, Also we are allowing user to set any valid logrus level not just debug or info level.

abhis2112: ok that make sense. Sure i can approve and merge now

abhis2112 (approved):

PR #1184, opened 2023-10-23
Set serviceBdRoutingDisable of fvBD to yes

akhilamohanan:
Set serviceBdRoutingDisable of fvBD to yes while creating service BD in aci-containers-controller

### DIFF
--- a/pkg/controller/services.go
+++ b/pkg/controller/services.go
@@ -235,6 +235,9 @@ func (cont *AciController) staticServiceObjs() apicapi.ApicSlice {
 	// Service bridge domain
 	bdName := cont.aciNameForKey("bd", cont.env.ServiceBd())
 	bd := apicapi.NewFvBD(cont.config.AciVrfTenant, bdName)
+	if apicapi.ApicVersion >= "6.0(3.84a)" {
+		bd.SetAttr("serviceBdRoutingDisable", "yes")
+	}
 	bd.SetAttr("arpFlood", "yes")
 	bd.SetAttr("ipLearning", "no")
 	bd.SetAttr("unkMacUcastAct", "flood")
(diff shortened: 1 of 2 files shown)


akhilamohanan:
Testing pending

snaiksat (approved):

pkg/controller/services.go
```diff
@@ -237,6 +237,7 @@ func (cont *AciController) staticServiceObjs() apicapi.ApicSlice {
 	bd := apicapi.NewFvBD(cont.config.AciVrfTenant, bdName)
 	bd.SetAttr("arpFlood", "yes")
 	bd.SetAttr("ipLearning", "no")
+	bd.SetAttr("serviceBdRoutingDisable", "yes")
```
snaiksat, line 240: what happens in the case of older ACI releases where this property is not supported?

snaiksat:
@akhilamohanan can you please close loop with @shastrinator , he shares my concern that we are not selectively performing this configuration for supported versions.

pkg/controller/services.go
```diff
@@ -235,6 +235,9 @@ func (cont *AciController) staticServiceObjs() apicapi.ApicSlice {
 	// Service bridge domain
 	bdName := cont.aciNameForKey("bd", cont.env.ServiceBd())
 	bd := apicapi.NewFvBD(cont.config.AciVrfTenant, bdName)
+	if apicapi.ApicVersion >= "6.0(3.77a)" {
```
snaiksat, line 238: I think 6.0(3.77a) is a Q/A release in preparation for 6.0.4. Assuming we want to only support this for 6.0(4x) and above, the comparison should be with a version number like this: 6.0(4x) as done here: https://github.com/noironetworks/aci-containers/blob/master/pkg/controller/controller.go#L676

Unfortunately, we don't know what "x" will be yet, and we will only know once its released. So this version comparison is tricky.

shastrinator: Hence it is safe to compare against the known version on which it is supported. It will continue
 to work for 6.04 when released.

snaiksat: okay, so we can keep the way it is

PR #1185, opened 2023-10-23
Support bridge-cni

shastrinator:
Bridge-cni with trunking support.
Tested with kube-virt


shastrinator:
Subsumed by https://github.com/noironetworks/aci-containers/pull/1216

PR #1197, opened 2023-11-16
Delete opflex-agent-cni and netop-cni before copying

akhilamohanan:
Delete opflex-agent-cni and netop-cni from CNIBIN when hostagent container comes up

### DIFF
--- a/docker/launch-hostagent.sh
+++ b/docker/launch-hostagent.sh
@@ -14,10 +14,11 @@ KUBECONFIG=/usr/local/etc/kubeconfig
 if [ -w /mnt/cni-bin ]; then
     # Install CNI plugin binary
     mkdir -p ${CNIBIN}
-    rm -f ${CNIBIN}/*
 if [ -z != $CHAINED_MODE ] && [ "$CHAINED_MODE" == "true" ]; then
+    rm -f ${CNIBIN}/netop-cni
     cp ${ACIBIN}/netop-cni $CNIBIN
 fi
+    rm -f ${CNIBIN}/opflex-agent-cni
     cp ${ACIBIN}/opflex-agent-cni $CNIBIN
 fi
 


docker/launch-hostagent.sh
```diff
@@ -14,10 +14,11 @@ KUBECONFIG=/usr/local/etc/kubeconfig
 if [ -w /mnt/cni-bin ]; then
     # Install CNI plugin binary
     mkdir -p ${CNIBIN}
-    rm -f ${CNIBIN}/*
 if [ -z != $CHAINED_MODE ] && [ "$CHAINED_MODE" == "true" ]; then
```
mchalla, line 17: Trying to understand why this if condition was not met with the previous change.

mchalla (approved):
Changes look fine. But pl see the comment.

docker/launch-hostagent.sh
akhilamohanan: This condition is for chained-mode setups. The issue is with rke setups where loopback binary will be present in CNIBIN folder and since we were deleting all binaries inside CNIBIN, loopback binary was also getting deleted (and not added back).

mchalla: Oh ok. I did not realize there were other binaries in there. Who copies them there ?

akhilamohanan: Initcontainer (cnideploy) container of hostagent pod copies it https://github.com/noironetworks/aci-containers/blob/master/docker/launch-cnideploy.sh

mchalla: Thanks

snaiksat: The idea is to delete the specific binary, not everything. It applies in both conditions, L19 as well as L21. We would hit L19 in non-ACI-CNI case.

snaiksat (approved):

PR #1201, opened 2023-11-21
Add webhook to append netop-cni to NADs

shastrinator:
Adds two containers: aci-containers-webhook and aci-containers-certmanager. Webhook handlers mutate network-attachment-definitions to insert netop-cni section. Webhook handlers can also be used for other resources following pattern here.
Cert-manager is added to ensure rotation of webhookcerts and also acts as CA for webhook. Platforms like openshift have their own CA and they can be used instead, provided rotation is ensured.


snaiksat (approved):

snaiksat:
thanks @shastrinator . Since the Docker* and lauch* files are identical in docker/ and docker/travis/* let's create a symlink from the docker/travis/* files to the docker/* files in a follow-up PR.

PR #1221, opened 2024-01-02
Support to push contract definition and contracts to EPG mapping from K8s into ACI

pkharat:


pkg/controller/nodefabricnetworkattachments.go
```diff
@@ -532,6 +535,71 @@ func (cont *AciController) applyNodeFabNetAttObjLocked(vlans []int, addNet *Addi
...
+		if tenantName != cont.config.AciPolicyTenant {
+			cont.log.Warningf("Different tenant '%s' provided in NetworkFabricConfiguration Resource, Ignoring it and using AciPolicyTenant '%s'", tenantName, cont.config.AciPolicyTenant)
+			tenantName = cont.config.AciPolicyTenant
+		}
+		commonTenantDn := "uni/tn-common"
+		epgObj := netFabConfig.Spec.VlanRefs[0].Epg
+		bdObj := epgObj.BD
+		vrfObj := bdObj.Vrf
+
+		// create BD
```
shastrinator, line 554: Do not create BD and EPG here. This is part of the function createNodeFabNetAttBd. Conditionally you have to either use the default prefix or the names in the CR. Contract should be a separate object with ref counts, do not create per nodefabnetatt object. This whole patch should be refactored.

shastrinator (changes requested):
See review comment.

shastrinator (approved):

PR #1234, opened 2024-01-18
Add pod-subet and node ips as remoteIp in hpp

akhilamohanan:
Add pod-subet and node ips as remoteIp in hpp for the networkpolicies which allow traffic from all pods in all namespaces, ie for the following cases:
	ingress:
	- from:
	  - namespaceSelector: {} podSelector: {}

	ingress:
	- from:
	  - namespaceSelector: {}

### DIFF
--- a/pkg/controller/config.go
+++ b/pkg/controller/config.go
@@ -181,7 +181,7 @@ type ControllerConfig struct {
 
 	// Pod subnet CIDRs in the form <gateway-address>/<prefix-length> that
 	// cover all pod-ip-pools
-	PodSubnets []string `json:"pod-subnets,omitempty"`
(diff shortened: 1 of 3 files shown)


pkg/controller/config.go
```diff
@@ -181,7 +181,7 @@ type ControllerConfig struct {
 
 	// Pod subnet CIDRs in the form <gateway-address>/<prefix-length> that
 	// cover all pod-ip-pools
-	PodSubnets []string `json:"pod-subnets,omitempty"`
+	PodSubnet []string `json:"pod-subnet,omitempty"`
```
snaiksat, line 184: why is this name changed required?

snaiksat (approved):

PR #1236, opened 2024-01-19
Disable HPP rendering based on disable-hpp-rendering flag

smshareef:


pkg/controller/environment.go
```diff
@@ -330,6 +334,12 @@ func (env *K8sEnvironment) PrepareRun(stopCh <-chan struct{}) error {
 			cont.deploymentInformer.HasSynced,
 			cont.podInformer.HasSynced,
 			cont.networkPolicyInformer.HasSynced)
+	} else if !cont.config.DisableHppRendering {
```
akhilamohanan, line 337: check for DisableHppRendering should be inside if !cont.config.ChainedMode {} . Otherwise it will wait for networkpolicy sync whenever chained mode is disbaled, execution will not go to the else if part

smshareef: Done, thank you @akhilamohanan.

PR #1239, opened 2024-01-26
Check for stale ports and start fresh if present

mchalla:
- stale ports can show up on reboot because we do not have any reboot cleanup triggers. This causes extra work for host-agent pod and by cleaning up database we can avoid the extra work
- the host-agent is resiliant enough that a sync is not needed. so say if host-ag came up first and added anything it will be redone. However we can use the same reboot.conf.d logic used in opflex for host-agent since it has access to the same file structure.


snaiksat (approved):
Thanks @mchalla . @akhilamohanan  let's target a follow-up patch to restart host-agent after OVS is up (similar to what we do for opflex-agent).

PR #1260, opened 2024-02-09
Capture LLDP using rawsockets for OVS ports

shastrinator:
Support for uplink port added to unmanaged OVS.
Requires CGO_ENABLED and libpcap on host.


shastrinator:
Superceded by https://github.com/noironetworks/aci-containers/pull/1261

PR #1261, opened 2024-02-14
Capture LLDP using rawsockets for OVS ports

shastrinator:
Support for uplink port added to unmanaged OVS.
Adds another build target aci-containers-host-ovscni, which requires CGO_ENABLED and libpcap required.
Corresponding changes in CICD.


docker/travis/Dockerfile-host
```diff
@@ -47,8 +47,15 @@ RUN alternatives --install /usr/sbin/iptables iptables /usr/sbin/iptables-legacy
...
 RUN echo "send dhcp-client-identifier = hardware;" > /usr/local/etc/dhclient.conf
 ENV TENANT=kube
 ENV NODE_EPG='kubernetes|kube-nodes'
+
+FROM base as with-ovscni
+COPY dist-static/aci-containers-host-agent-ovscni dist-static/opflex-agent-cni dist-static/netop-cni docker/launch-hostagent.sh docker/enable-hostacc.sh docker/enable-droplog.sh /usr/local/bin/
+RUN mv /usr/local/bin/aci-containers-host-agent-ovscni /usr/local/bin/aci-containers-host-agent
+CMD ["/usr/local/bin/launch-hostagent.sh"]
+
+FROM base as without-ovscni
```
snaiksat, line 59: since this is default, should we get rid of the target? I understand that by adding the target we get both, default as well as explicit-with-the-target but just thinking from a readability perspective which is better...

snaiksat (approved):

PR #1267, opened 2024-02-20
Periodically check and remove the taint added by the controller

smshareef:


pkg/hostagent/agent.go
```diff
@@ -497,22 +478,35 @@ func (agent *HostAgent) runTickers(stopCh <-chan struct{}) {
...
 			}
-			agent.taintRemoved.Store(true)
+			for _, taint := range node.Spec.Taints {
+				if taint.Key == "aci-containers-host/unavailable" {
+					agent.indexMutex.Lock()
+					count := len(agent.completedSyncTypes)
+					agent.indexMutex.Unlock()
+
+					if count == 5 {
+						err := agent.removeTaint(node, "aci-containers-host/unavailable", kubeClient)
```
akhilamohanan, line 500: can remove taint only if node is ready?

smshareef: Yes, we can do that.

pkg/hostagent/agent.go
```diff
@@ -497,22 +478,35 @@ func (agent *HostAgent) runTickers(stopCh <-chan struct{}) {
...
+		case <-ticker.C:
+			node, err := kubeClient.CoreV1().Nodes().Get(context.TODO(), os.Getenv("KUBERNETES_NODE_NAME"), metav1.GetOptions{})
 			if err != nil {
-				agent.log.Errorf("Failed to remove taint: %v", err)
 				continue
 			}
-			agent.taintRemoved.Store(true)
+			for _, taint := range node.Spec.Taints {
+				if taint.Key == "aci-containers-host/unavailable" {
+					agent.indexMutex.Lock()
```
PariyaAshok, line 495: I think check should based on key as well as effect, We might have same key with different tolerations effect,
Also can we avoid hardcoded key value here?

smshareef: Thank you for the review Ashok. Made the changes.

pkg/hostagent/agent.go
smshareef: Done Akhila, thanks.

pkg/controller/nodes.go
```diff
@@ -47,6 +47,11 @@ const (
 	tunnelIDIncr = 2
 )
 
+// Name of the taint to add to nodes that are not ready
+const (
+	ACIContainersTaintName string = "aci-containers-host/unavailable"
+)
+
```
PariyaAshok, line 54: @smshareef  How are we making sure if we are removing taint only added by controller?
With this fix key-name we will remove taint not set by controller right? Can we generate taint-key(key=value:effect) dynamically and set it??

smshareef: @PariyaAshok Yes, if the taint with this key and effect NoSchedule is added, then we will remove. We can document that the taint with this key is managed by the ACI-CNI.

If we generate the taint key dynamically, then the same needs to be communicated between the controller and host-agent.

snaiksat (approved):

PR #1273, opened 2024-02-23
Made istio.io vendor package a part of aci-containers packages

akhilamohanan:
Grype tool was detecting vulnerabilities for aci-containers-controller image due to the version mismatch in istio go package which was using pseudo version and istio github tags which doesn't have v in it

Made istio.io vendor package a part of aci-containers packages to avoid the issue

Made following changes as part of PR:
* Moved all the files inside istio.io package to two separate folders in aci-containers packages
* Modified the package name where ever necessary
* Used the modified packages in aci-containers where ever istio.io package was referred 
* Above steps removed dependency of istio.io package


akhilamohanan:
Handled the issue using a different approach https://github.com/noironetworks/aci-containers/pull/1274

PR #1274, opened 2024-02-23
Removed istio.io/istio package dependency

akhilamohanan:
quay.io security scan was identifying vulnerabilities for istio.io/istio package. Removed the package dependency by commenting out the code that reffered it.


abhis2112 (changes requested):
Request changes

abhis2112 (approved):

PR #1279, opened 2024-03-01
Fix for processing networkpolicy to allow specific protocol traffic

akhilamohanan:
Networkpolicy to allow any port traffic to a server on a specific protocol was failing as rule was not created for a networkpolicy with no port number mentioned, but has protocol.

Modified code to add rule for a networkpolicy with protocol without port.


PariyaAshok (approved):
lgtm.

snaiksat (approved):

PR #1286, opened 2024-03-15
Do VLAN programming upfront for OpenShift on OpenStack

akhilamohanan:
For OpenShift on OpenStack clusters, vnsRsCIfPathAtt is created for each OpenStack compute hosts instead of OpenShift nodes so that VLAN will already be there when a vm is migrated from one host to other.


akhilamohanan:
closing the PR as it is handled in https://github.com/noironetworks/aci-containers/pull/1287

PR #1287, opened 2024-03-18
Do VLAN programming upfront for OpenShift on OpenStack

akhilamohanan:
For OpenShift on OpenStack clusters, vnsRsCIfPathAtt is created for each OpenStack compute hosts along with OpenShift nodes so that VLAN will already be there when a vm is migrated from one compute host to other.

### DIFF
--- a/pkg/controller/services.go
+++ b/pkg/controller/services.go
@@ -1277,6 +1277,12 @@ func (cont *AciController) updateDeviceCluster() {
 		}
 		nodeMap[node] = fabricPath
 	}
+
+	// For clusters other than OpenShift On OpenStack,
+	// openStackFabricPathDnMap will be empty
+	for host, opflexOdevInfo := range cont.openStackFabricPathDnMap {
+		nodeMap[host] = opflexOdevInfo.fabricPathDn
+	}
 	cont.indexMutex.Unlock()
 
 	var nodes []string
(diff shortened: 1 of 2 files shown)


pkg/controller/services.go
```diff
@@ -1321,6 +1327,23 @@ func (cont *AciController) opflexDeviceChanged(obj apicapi.ApicObject) {
...
+				opflexOdevInfo.count++
+				cont.openStackFabricPathDnMap[obj.GetAttrStr("hostName")] = opflexOdevInfo
+			} else {
+				var openstackopflexodevinfo openstackOpflexOdevInfo
+				openstackopflexodevinfo.count++
+				openstackopflexodevinfo.fabricPathDn = obj.GetAttrStr("fabricPathDn")
+				openstackopflexodevinfo.dn = obj.GetAttrStr("dn")
+				cont.openStackFabricPathDnMap[obj.GetAttrStr("hostName")] = openstackopflexodevinfo
+			}
+			cont.updateDeviceCluster()
```
snaiksat, line 1344: given that we dont anticipate too many opflex add events for the openstack nodes, should we log here at info level to indicate that we are processing at this point?

akhilamohanan: Added log

snaiksat (approved):

PR #1299, opened 2024-04-02
Only use newer upstream sriovnet dep

tomflynn:


snaiksat:
thanks @tomflynn , we have requested @Bhavanaashok33 to test this out if a setup is available

Bhavanaashok33 (approved):

PR #1308, opened 2024-05-08
Update topology aware routing labels

tomflynn:
Look for topology.kubernetes.io/zone label not kubernetes.io/zone on Node resource
service.kubernetes.io/topology-mode is the service annotation as of 1.27

Use upstream label key constants


shastrinator (approved):
LGTM. Yes zone annotation was incorrect. Thanks.

PR #1318, opened 2024-05-31
Adding changes related to aci cilium cni chaining

abhis2112:
Changes made cater to the request of generating 10-opflex.conf file if we use generic_veth mode for cilium cni chaining.


shastrinator (approved):
LGTM.

PR #1343, opened 2024-07-17
Added webook, certmanager UTs in make check

akhilamohanan:
* Added webook, certmanager UTs in make check
* Added webook, certmanager and aci-containers-operator in goinstall of Makefile

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -156,6 +156,9 @@ goinstall:
 	${INSTALL_CMD} ${BASE}/cmd/controller
 	${INSTALL_CMD} ${BASE}/cmd/hostagent
 	${INSTALL_CMD} ${BASE}/cmd/acikubectl
+	${INSTALL_CMD} ${BASE}/cmd/acicontainersoperator
+	${INSTALL_CMD} ${BASE}/cmd/webhook
+	${INSTALL_CMD} ${BASE}/cmd/certmanager
 
 dist/opflex-agent-cni: ${AGENTCNI_DEPS}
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -268,7 +271,7 @@ fix-gofmt:
 lint:
 	@${LINT_CMD} run
 
-check: check-gofmt check-ipam check-index check-apicapi check-controller check-hostagent check-gbpserver
+check: check-gofmt check-ipam check-index check-apicapi check-controller check-hostagent check-gbpserver check-webhook check-certmanager
```
snaiksat, line 274: do we need for acicontainersoperator as well?

exclude-covprof.conf
```diff
@@ -21,3 +21,6 @@ github.com/noironetworks/aci-containers/pkg/hostagent/netattachdef.go
 github.com/noironetworks/aci-containers/pkg/hostagent/pod_relatives.go
 github.com/noironetworks/aci-containers/pkg/hostagent/setup.go
 github.com/noironetworks/aci-containers/pkg/hostagent/snatlocalinfo.go
+github.com/noironetworks/aci-containers/pkg/webhook
+github.com/noironetworks/aci-containers/cmd/webhook
```
snaiksat, line 25: this appears twice...

akhilamohanan: one is for cmd and other is for pkg

Makefile
akhilamohanan: addressed

snaiksat (approved):

PR #1372, opened 2024-08-27
Fixed ipRouteP, ipNexthopP MO creation failure

YogeshRajmane:


shastrinator (approved):
Thanks for catching this.

PR #1415, opened 2024-11-14
Increase subscription refresh offset

shastrinator:
Atleast with lldpIf MO, subscription expiry seems to happen atleast 190 secs earlier than expected, so
increasing the offset to 210 seconds.


snaiksat (approved):

snaiksat:
Thanks, please also backport to the mmr-6.1.1 branch.

PR #1463, opened 2024-12-06
Added flavors for OCP-4.17 Baremetal, ESX and Openstack

burhan20:


pkg/hostagent/services.go
```diff
@@ -119,6 +119,7 @@ var Version = map[string]bool{
 	"openshift-4.14-baremetal":             true,
 	"openshift-4.15-baremetal":             true,
 	"openshift-4.16-baremetal":             true,
+	"openshift-4.17-baremetal":             true,
 	"openshift-4.14-agent-based-baremetal": true,
 	"openshift-4.15-agent-based-baremetal": true,
 	"openshift-4.16-agent-based-baremetal": true,
```
snaiksat, line 125: we need to add here as well, right?

abhis2112 (approved):

PR #1467, opened 2024-12-09
update SnatGlobalInfo CR when uplink MAC address changes

akhilamohanan:
When uplink MAC address changes, the SnatGlobalInfo was not getting updated which was causing disruption in SNAT traffic


fwardzic (approved):
LGTM

PR #1525, opened 2025-05-21
Enable Resilient Hashing in LoadBalancer PBR

akhilamohanan:
Add change to enable Resilient Hashing in L4-L7 Redirection Policy by default, when exposing a service as type LoadBalancer. When this option is enabled, ACI fabric wont re-hashing existing flows when a L3 Destination will be deleted or added.

### DIFF
--- a/pkg/apicapi/apic_metadata.go
+++ b/pkg/apicapi/apic_metadata.go
@@ -635,8 +635,9 @@ var metadata = map[string]*apicMeta{
 	},
 	"vnsSvcRedirectPol": {
 		attributes: map[string]interface{}{
-			"name":                "",
-			"thresholdDownAction": "permit",
+			"name":                 "",
+			"thresholdDownAction":  "permit",
+			"resilientHashEnabled": "yes",
 		},
 		children: []string{
 			"vnsRedirectDest",
--- a/pkg/controller/config.go
+++ b/pkg/controller/config.go
@@ -139,6 +139,10 @@ type ControllerConfig struct {
 	// is enabled for snat
 	AciPbrTrackingNonSnat bool `json:"aci-pbr-tracking-non-snat,omitempty"`
 
+	// By default, the Resilient Hashing Enabled field of vnsSvcRedirectPol is
+	// set to "yes". If DisableResilientHashing is true, it will be set to "no"
+	DisableResilientHashing bool `json:"disable-resilient-hashing,omitempty"`
+
 	// The tenants related to AciVrf where BDs/EPGs/Subnets could exist.
 	// Usually AciVrfTenant and AciPolicyTenant
 	AciVrfRelatedTenants []string `json:"aci-vrf-related-tenants,omitempty"`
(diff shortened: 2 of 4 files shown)


pkg/controller/services.go
```diff
@@ -661,6 +661,9 @@ func (cont *AciController) apicRedirectPol(name string, tenantName string, nodes
 	monPolDn string, enablePbrTracking bool) (apicapi.ApicObject, string) {
 	rp := apicapi.NewVnsSvcRedirectPol(tenantName, name)
 	rp.SetAttr("thresholdDownAction", "deny")
+	if !cont.config.DisableResilientHashing {
+		rp.SetAttr("resilientHashEnabled", "yes")
+	}
```
snaiksat, line 666: If we always set to "yes" in:
https://github.com/noironetworks/aci-containers/pull/1525/files#diff-0b386e1ef0eec04bf5d5d93f0d5672149bf33dff5aadb7a3c7e9aa01d7934d3aR640

you can avoid the double negation here and say:
if cont.config.DisableResilientHashing {
		rp.SetAttr("resilientHashEnabled", "no")
	}

pkg/controller/config.go
```diff
@@ -139,6 +139,10 @@ type ControllerConfig struct {
 	// is enabled for snat
 	AciPbrTrackingNonSnat bool `json:"aci-pbr-tracking-non-snat,omitempty"`
 
+	// By default, the Resilient Hashing Enabled field of vnsSvcRedirectPol is
+	// set to "yes". If DisableResilientHashing is true, it will be set to "no"
+	DisableResilientHashing bool `json:"disable-resilient-hashing,omitempty"`
```
snaiksat, line 144: This should default to True if empty?

akhilamohanan: This will be false if empty

pkg/apicapi/apic_metadata.go
```diff
@@ -635,8 +635,9 @@ var metadata = map[string]*apicMeta{
 	},
 	"vnsSvcRedirectPol": {
 		attributes: map[string]interface{}{
-			"name":                "",
-			"thresholdDownAction": "permit",
+			"name":                 "",
+			"thresholdDownAction":  "permit",
+			"resilientHashEnabled": "no",
```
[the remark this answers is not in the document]
akhilamohanan: func NewVnsSvcRedirectPol(tenantName, name string) ApicObject {
        ret := newApicObject("vnsSvcRedirectPol")
        ret["vnsSvcRedirectPol"].Attributes["name"] = name
        ret["vnsSvcRedirectPol"].Attributes["dn"] =
                fmt.Sprintf("uni/tn-%s/svcCont/svcRedirectPol-%s", tenantName, name)
        return ret
}  
These are the 2 fields we set when we create NewVnsSvcRedirectPol. Everything else will be APIC default value (not default value in metadata) if we don't set explicitly

pkg/controller/services.go
akhilamohanan: addressed

PR #1526, opened 2025-05-21
Subscribe vpcIf for service VLAN pre-provisioning

akhilamohanan:
Subscribe to both infraRtAttEntP and vpcIf, and process vpcIf entries whose accBndlGrpDn matches the tDn of infraRtAttEntP. This supports cases where a new IPG is created and then the interfaces are assigned.

### DIFF
--- a/pkg/controller/services.go
+++ b/pkg/controller/services.go
@@ -1306,8 +1306,10 @@ func (cont *AciController) updateDeviceCluster() {
 
 	// For OpenShift On OpenStack clusters,
 	// hostFabricPathDnMap will be empty
-	for host, fabricPathDn := range cont.hostFabricPathDnMap {
-		nodeMap[host] = fabricPathDn
+	for _, hostInfo := range cont.hostFabricPathDnMap {
+		if hostInfo.fabricPathDn != "" {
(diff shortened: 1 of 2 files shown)


pkg/controller/services.go
```diff
@@ -1455,58 +1460,125 @@ func (cont *AciController) infraRtAttEntPDeleted(dn string) {
...
+	for tDn, hostInfo := range cont.hostFabricPathDnMap {
+		if _, present := hostInfo.vpcIfDn[dn]; present {
+			cont.log.Info("Deleting vpcIf, dn :", dn)
+			delete(hostInfo.vpcIfDn, dn)
+			if len(hostInfo.vpcIfDn) == 0 {
+				cont.log.Infof("Removing fabricPathDn(%s) of ipg : %s ", hostInfo.fabricPathDn, hostInfo.host)
+				hostInfo.fabricPathDn = ""
+				deleted = true
+			}
+			cont.hostFabricPathDnMap[tDn] = hostInfo
```
shastrinator, line 1475: Nit: Shouldn't you delete this mapping itself at this point. As long as its handled in updateDeviceCluster, its ok.

shastrinator (approved):
Looks ok. Please let me know how the testing goes..

PR #1653, opened 2026-01-07
Add support for port range in network policy

allenantony-oc:
The controller previously ignored the endPort field in NetworkPolicy rules. This resulted in incorrect ACI hostprotRule objects that only allowed traffic to the first port of an intended range.

This commit updates the controller to correctly process endPort. It now maps the port field to fromPort and endPort to toPort, ensuring that port ranges defined in a NetworkPolicy are properly enforced in ACI.


maheshkurund (approved):
LGTM

PR #1667, opened 2026-02-13
Fix NetworkPolicy named port resolution in HPP modes

allenantony-oc:
When multiple NetworkPolicies share the same HPP and use named ports that resolve to different port numbers, only one policy's rules were generated in hpp-optimization and hpp-direct modes.

Root cause: HPP hash is based on ingress/egress rules and policy types. NetworkPolicies with identical rules but different named port mappings share the same HPP. The last NP's rules replace the previous NPs' rules causing unintended traffic disruption for previous NPs.

Changes:
- Iterate through all NetworkPolicies sharing an HPP when building ingress/egress rules in both hpp-optimization and hpp-direct modes
- Add NetworkPolicy name suffix to rules in buildLocalNetPolSubjRules to ensure unique rule names and match buildNetPolSubjRules behavior

This ensures all named port mappings are included when multiple NPs share an HPP, preventing traffic disruption.

### DIFF
--- a/pkg/controller/network_policy.go
+++ b/pkg/controller/network_policy.go
@@ -1371,11 +1371,15 @@ func (cont *AciController) buildLocalNetPolSubjRules(ruleName string,
 	logger *logrus.Entry, npKey string, np *v1net.NetworkPolicy, peerIpBlock []string) {
 	if len(ports) == 0 {
 		if !cont.configuredPodNetworkIps.V4.Empty() {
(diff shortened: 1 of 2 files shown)


allenantony-dev:
~~closing the PR as it is handled in https://github.com/noironetworks/aci-containers/pull/1707~~

jeffinkottaram:
Handled in https://github.com/noironetworks/aci-containers/pull/1734

PR #1687, opened 2026-03-13
Add infra querier subnet handling for multipod migration

bravi64:
In a multipod ACI fabric, IGMP queries on the infra VLAN originate
from the Pod 1 TEP pool subnet (e.g. 10.0.0.30). VMs outside Pod 1
drop these queries due to reverse-path filtering — they have no route
to the source subnet, so the RPF check fails and IGMP reports are
never sent back. This breaks multicast group membership for migrated
VMs.

To address this, a static route for the infra querier subnet is added
on the infra VLAN interface — the same approach already used for the
224.0.0.0/4 multicast route. This commit handles the case where a
DHCP release-renew cycle during inter-pod migration clears the
existing routes.

Controller:
- Queries APIC for the querier subnet under tn-infra/BD-default on
  startup and on every APIC reconnect
- Propagates the subnet to nodes via a K8s annotation, updating only
  nodes that carry a stale or missing value

Hostagent:
- Reads the annotation and adds the route after each DHCP renew cycle
- Waits with exponential backoff if the annotation has not yet
  propagated, giving up after approximately five minutes
- Uses netlink for route management instead of shelling out
- Deduplicates concurrent route additions and detects existing
  superset routes to avoid redundant entries

### DIFF
--- a/pkg/hostagent/opflex.go
+++ b/pkg/hostagent/opflex.go
@@ -27,6 +27,7 @@ import (
 	"reflect"
 	"regexp"
 	"strings"
+	"syscall"
 	"text/template"
 	"time"
 
@@ -42,6 +43,7 @@ const (
 	PLATFORM_SUBSCRIPTION_FILE = "/usr/local/var/lib/opflex-agent-ovs/events/platformconfig.subscriptions"
 	PLATFORM_NOTIFICATION_FILE = "/usr/local/var/lib/opflex-agent-ovs/events/platformconfig.notifications"
 	MAX_PLATFORM_EVENT_AGE     = 120 * time.Second
+	infraRouteMetric           = 401
 )
 
 type opflexFault struct {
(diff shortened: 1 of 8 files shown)


pkg/hostagent/opflex.go
```diff
@@ -187,28 +188,114 @@ func (agent *HostAgent) isMultiCastRoutePresent(link netlink.Link) bool {
...
+			subnet, _ := agent.infraQuerierSubnet.Load().(string)
+			if subnet != "" {
+				agent.addInfraQuerierSubnetRoute(link, name, subnet)
+				return
+			}
+			agent.log.Infof("Infra querier subnet not yet available, retrying in %v", backoff)
+			time.Sleep(backoff)
+			backoff *= 2
+		}
+		agent.log.Error("Infra querier subnet not available after retries, giving up for interface ", name)
```
shchatur-us: Not sure what this means. Could you please provide more clarity in the error message?

pkg/hostagent/opflex.go
```diff
@@ -187,28 +188,114 @@ func (agent *HostAgent) isMultiCastRoutePresent(link netlink.Link) bool {
...
+// (5→10→20→40→80→160s ≈ 5 min total), then gives up. Once the subnet is known, route
+// addition follows the same bounded-retry pattern as addMultiCastRoute.
+func (agent *HostAgent) scheduleInfraQuerierSubnetRoute(link netlink.Link, name string) {
+	if !agent.infraRoutePending.CompareAndSwap(false, true) {
+		agent.log.Info("Infra querier route addition already pending for ", name)
+		return
 	}
+	go func() {
+		defer agent.infraRoutePending.Store(false)
+		backoff := 5 * time.Second
```
shchatur-us: Consider defining and using constants here and on the next line, instead of magic numbers.

pkg/hostagent/opflex.go
```diff
@@ -187,28 +188,114 @@ func (agent *HostAgent) isMultiCastRoutePresent(link netlink.Link) bool {
...
+		for range 6 {
+			subnet, _ := agent.infraQuerierSubnet.Load().(string)
+			if subnet != "" {
+				agent.addInfraQuerierSubnetRoute(link, name, subnet)
+				return
+			}
+			agent.log.Infof("Infra querier subnet not yet available, retrying in %v", backoff)
+			time.Sleep(backoff)
+			backoff *= 2
```
shchatur-us: Same as above.

pkg/hostagent/opflex.go
```diff
@@ -187,28 +188,114 @@ func (agent *HostAgent) isMultiCastRoutePresent(link netlink.Link) bool {
...
+	}
+
+	if agent.isInfraQuerierSubnetRoutePresent(link, subnetNet) {
+		agent.log.Info("Infra querier subnet route already present for interface ", name)
+		return
+	}
+
+	retryCount := agent.config.DhcpRenewMaxRetryCount
+	for i := 0; i < retryCount; i++ {
+		if err := agent.addStaticLinkRoute(link, subnetNet, 401); err != nil {
```
shchatur-us: Could there be other errors than 401?

pkg/hostagent/opflex.go
jeffinkottaram: Switched to a more detailed error message.

pkg/hostagent/opflex.go
jeffinkottaram: 401 here is the metric with which the route is added. Declared and used a variable for this as well in addition to changing the other magic numbers.

PR #1702, opened 2026-04-15
Remove stale binary dependencies from CNI images

jeffinkottaram:
Several pinned binaries in the ACI container images are outdated. Istio support was previously disabled due to known CVEs in istio.io/istio; the kubectl and istioctl installs in the controller images were the only remaining artefacts of that disabled code path. This change removes those stale installs and replaces the operator's kubectl dependency with a native Go implementation.

Operator

- Replace exec("kubectl apply") in handleOperatorCreate with native controller-runtime server-side apply via pkg/util/applyresources.go, using unstructured SSA
- Remove the aci-deployment.yaml file write round-trip, no longer needed
- Delete Dockerfile-operator-base and rebase Dockerfile-operator directly onto aci-containers-base
- Remove kubectl installation from dev/alpine and travis operator Dockerfiles

Controller

- Remove stale kubectl and istioctl installs from all controller Dockerfile variants; the Go code consuming them is already build-excluded
- Disable COPY of upstream-istio-cr.yaml in controller image

Cnideploy

- Bump CNI plugins from v0.9.1 to v1.9.1
- Parameterise with ARG CNI_PLUGINS_VERSION for easier future updates

### DIFF
--- a/docker/Dockerfile-cnideploy-base
+++ b/docker/Dockerfile-cnideploy-base
@@ -1,7 +1,8 @@
 ARG basetag=latest
 ARG baserepo=quay.io/noirolabs
 FROM ${baserepo}/aci-containers-base:${basetag}
+ARG CNI_PLUGINS_VERSION=v1.9.1
 RUN yum install -y wget ca-certificates tar gzip \
   && yum clean all \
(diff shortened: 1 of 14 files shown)


maheshkurund (approved):
LGTM

YogeshRajmane (approved):

PR #1769, opened 2026-08-15
ci: update cicd-status inline via push-to-cicd-status.sh; drop diverg…

anmol372:
…ent publisher

Provide TRAVIS_TAGGER (GITHUBACTIONS_TAGGER secret) to the build step so the un-gated push-to-cicd-status.sh can write to cicd-status main, and remove the separate publish-github-actions-test-status.sh preview step.


anmol372:
Closing with companion CICD #81. On its own this removes the active ACI status publisher; paired with #81 it replaces the merged atomic publisher with per-image updates while leaving OpFlex double-wired.

